PR #3, opened 2015-04-21
implement add-ons, error message, character counter

morethanreal:
@cdata @notwaldorf PTAL

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,7 +1,9 @@
 {
   "name": "paper-input",
   "version": "0.8.0",
-  "authors": "The Polymer Authors",
+  "authors": [
+    "The Polymer Authors"
+  ],
   "keywords": [
     "web-components",
     "web-component",
@@ -19,7 +21,8 @@
   "homepage": "https://github.com/PolymerElements/paper-input",
   "ignore": [],
   "dependencies": {
-    "polymer": "Polymer/polymer#v0.8.0-rc.6",
+    "polymer": "Polymer/polymer#v0.8.0-rc.7",
+    "iron-input": "PolymerElements/iron-input#^0.8.0",
     "paper-styles": "PolymerElements/paper-styles#^0.8.0"
   },
   "devDependencies": {
--- a/demo/index.html
+++ b/demo/index.html
@@ -19,7 +19,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   <script src="../../webcomponentsjs/webcomponents-lite.js"></script>
 
+  <link rel="import" href="../paper-input.html">
   <link rel="import" href="../paper-input-container.html">
+  <link rel="import" href="../paper-input-error.html">
+  <link rel="import" href="../paper-input-char-counter.html">
+  <link rel="import" href="../../iron-input/iron-input.html">
 
   <link rel="stylesheet" href="../../paper-styles/demo.css">
 
@@ -28,14 +32,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   <section>
 
+    <paper-input label="label"></paper-input>
+
+    <paper-input no-label-float label="label (no-label-float)"></paper-input>
+
     <paper-input-container>
       <label>label</label>
-      <input>
-    </paper-input-container>
-
-    <paper-input-container no-label-float>
-      <label>label (no-label-float)</label>
-      <input>
+      <textarea></textarea>
     </paper-input-container>
 
   </section>
@@ -44,10 +47,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     <div>Auto Validation</div>
 
-    <paper-input-container auto-validate>
-      <label>letters only</label>
-      <input pattern="[a-zA-Z]*">
-    </paper-input-container>
+    <paper-input label="only type letters" auto-validate pattern="[a-zA-Z]*" value="123" error-message="letters only!"></paper-input>
+
+  </section>
+
+  <section>
+
+    <div>Character Counter</div>
+
+    <paper-input label="label" char-counter></paper-input>
+
+    <paper-input label="at most 5 characters" char-counter maxlength="5"></paper-input>
 
   </section>
 
--- a/paper-input-char-counter.html
+++ b/paper-input-char-counter.html
@@ -1,0 +1,96 @@
+<!--
+@license
+Copyright (c) 2015 The Polymer Project Authors. All rights reserved.
+This code may only be used under the BSD style license found at http://polymer.github.io/LICENSE.txt
+The complete set of authors may be found at http://polymer.github.io/AUTHORS.txt
+The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+Code distributed by Google as part of the polymer project is also
+subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+-->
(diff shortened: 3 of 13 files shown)


paper-input-container.html
```diff
@@ -266,13 +302,28 @@
...
       }
       if (this.autoValidate) {
-        this._inputIsInvalid = !event.target.validity.valid;
+        this._inputIsInvalid = !inputElement.validity.valid;
+      }
+      // notify add-ons
+      var addons = Polymer.dom(this.$.addOnContent).getDistributedNodes();
+      for (var a, i = 0; a = addons[i]; i++) {
+        // this could get called before the distributed node is upgraded
+        a.fire && a.fire('input-value-changed', {inputElement: inputElement, value: value});
```
cdata: Manually triggering an event on other elements seems like a very manual means to achieve the same end as binding. Is it possible to make this work with a binding instead?

demo/index.html
```diff
@@ -45,8 +48,27 @@
     <div>Auto Validation</div>
 
     <paper-input-container auto-validate>
-      <label>letters only</label>
-      <input pattern="[a-zA-Z]*">
+      <label>only type letters</label>
+      <input is="iron-input" pattern="[a-zA-Z]*">
+      <paper-input-error class="add-on">letters only!</paper-input-error>
```
cdata: In the case of buttons, we ship a compound widget (for example, `paper-icon-button`) composed of several component parts (card, ripple, icon) and optimize for ease of putting the element in the document (you just add a single tag, and configure with attributes).

In this case, we ask the user to configure a compound widget by composing many atomic widgets in their local DOM. This optimizes for modularity and configuration, but is less easy to add to the document.

This seems like a double standard. Both strategies have merit, but as a user of a package of elements, I probably only expect one strategy or another for that entire package. Which of the two strategies is the "correct" way forward for ours?

paper-input-container.html
```diff
@@ -239,22 +266,31 @@
       'input': '_onInput'
     },
 
+    get _valueChangedEvent() {
+      return this.attrForValue + '-changed';
+    },
+
+    get _propertyForValue() {
+      return Polymer.CaseMap.dashToCamelCase(this.attrForValue);
```
cdata: :bulb: I just learned about something new here.

paper-input-container.html
```diff
@@ -239,22 +266,31 @@
...
+    },
+
     attached: function() {
       var input = Polymer.dom(this).querySelector('input,.paper-input-input');
-      this._inputHasContent = !!input.value;
-      this._inputIsInvalid = !input.validity.valid;
+      this._handleInput(input);
+      // once more for add-ons, async to wait for distribution
```
cdata: What is the distribution pitfall encountered here?

paper-input-container.html
morethanreal: I'm not sure there's an approved way to manually set up a binding outside of the element's template, and that also seems difficult to decipher. Since this is a distributed node an event seem to be the most standard method for communication.

paper-input-container.html
morethanreal: I think since the elements are made ready top-down, the child nodes aren't ready at this time to get the event in `_handleInput`.

paper-input-container.html
notwaldorf: Whoaaaaaa same. :balloon:

demo/index.html
morethanreal: We should absolutely have pre-composed elements for ease of use. For this element I was thinking about giving the option of composing your own input element if you need more flexibility, and also providing pre-composed elements for convenience. I have not yet made the convenience elements :)

The API surface for `input` is large, so it seems impractical to mirror it in entirety if we only provide a pre-composed element.

paper-input-container.html
morethanreal: What if I set properties on the other elements instead?

cdata: I trust your intuition. When I first saw this, it was from the receiving node's perspective, and from there it looked like magic ("where is this `input-value-changed` event even coming from?").

demo/index.html
cdata: :+1:

morethanreal:
Many changes, PTAnotherL? @cdata @notwaldorf

bower.json
```diff
@@ -19,8 +19,9 @@
   "homepage": "https://github.com/PolymerElements/paper-input",
   "ignore": [],
   "dependencies": {
-    "polymer": "Polymer/polymer#v0.8.0-rc.6",
-    "paper-styles": "PolymerElements/paper-styles#^0.8.0"
+    "iron-input": "PolymerElements/iron-input#^0.8.0",
+    "paper-styles": "PolymerElements/paper-styles#^0.8.0",
+    "polymer": "Polymer/polymer#v0.8.0-rc.7"
```
cdata: Polymer is the best dependency and deserves to go first (jk only sort of not jk).

paper-input-char-counter.html
```diff
@@ -0,0 +1,96 @@
...
+
+<script>
+
+(function() {
+
+  Polymer({
+
+    is: 'paper-input-char-counter',
+
+    enableCustomStyleProperties: true,
```
cdata: I wish this would just be a default thing. My guess is that it will get deprecated but we'll still be removing vestigial `enableCustomStyleProperties` for years to come.

paper-input-char-counter.html
```diff
@@ -0,0 +1,96 @@
...
+
+    hostAttributes: {
+      'class': 'add-on'
```
cdata: Is it safe to set `class` with `hostAttributes`?

paper-input-char-counter.html
morethanreal: Yeah, @sorvell promises to remove it soon.

paper-input-char-counter.html
morethanreal: I'm using a class for `<content select>` right now. Maybe change to use an attribute so the user can't stomp on it?

paper-input-error.html
```diff
@@ -0,0 +1,102 @@
...
+(function() {
+
+  Polymer({
+
+    is: 'paper-input-error',
+
+    enableCustomStyleProperties: true,
+
+    hostAttributes: {
+      'class': 'add-on',
```
cdata: Ditto re: https://github.com/PolymerElements/paper-input/pull/3/files#r28933170

paper-input.html
```diff
@@ -7,4 +7,123 @@
...
+`<paper-input>` is a text field.
+-->
+
+<dom-module id="paper-input">
+
+  <template>
+
+    <paper-input-container no-label-float$="[[noLabelFloat]]" auto-validate$="[[autoValidate]]">
+
+      <template is="x-if" if="[[label]]">
```
cdata: Neat :candy:

paper-input-char-counter.html
cdata: Yeah, I think that's a safer choice.

cdata:
This is looking really cool. Minor details to address, but otherwise please :shipit:

bower.json
morethanreal: Haha. fixed.

paper-input-char-counter.html
morethanreal: Fixed.

morethanreal:
Squashing everything now.

PR #4, opened 2015-04-23
add type attribute

morethanreal:
@cdata @notwaldorf PTAL

I don't have a test, but there seems to be no way to send key events to the input and have it alter the value :(


cdata:
Nice, LGTM :+1:

notwaldorf:
lgtm also! :ship:

PR #6, opened 2015-04-23
fix error not showing with type="number"

morethanreal:
PTAL @cdata @notwaldorf

### DIFF
--- a/paper-input-error.html
+++ b/paper-input-error.html
@@ -31,7 +31,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       mixin(--paper-input-error);
     }
 
-    :host(.is-invalid) {
+    :host([invalid]) {
       display: inline-block;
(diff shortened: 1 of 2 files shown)


paper-input-error.html
```diff
@@ -63,19 +63,11 @@
...
-        type: Object
-      },
-
-      /**
-       * The current value of the input element.
-       */
-      value: {
-        observer: '_valueChanged',
-        type: String
+      invalid: {
```
cdata: Just an idea, but this attribute falls into the category of attributes I wish were reflected when I'm trying to debug things by scanning the Elements tab in inspector.

cdata:
LGTM :+1:

paper-input-error.html
morethanreal: I'll just make it `reflectToAttribute:true` then?

PR #8, opened 2015-04-24
Add a 'required' attribute to paper-input.

jklein24:


paper-input.html
```diff
@@ -90,6 +90,14 @@
       },
 
       /**
+       * Marks the `input` as required.
```
morethanreal: How about "set to true to mark the `input` as required" to match the other doc string?

morethanreal:
lgtm w/ nit

morethanreal:
Please fix the nit and squash the commits.

jklein24:
Done.

Issue #10, opened 2015-04-30
Label out of the input


LM450N:
I tried the `paper-input` element on Chrome 42 on Linux and the labels are all at the top left corner.

![label-out-of-input](https://cloud.githubusercontent.com/assets/2325091/7419698/29ebf430-ef77-11e4-87e6-0072b1e4954c.png)

notwaldorf:
Eeeek I'll fix it right now. It's from yesterday's change to `iron-flex-layout`. Sorry!

PR #12, opened 2015-04-30
fix layout classes

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/10

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -166,6 +166,11 @@ For example:
       color: var(--paper-input-container-focus-color);
     }
 
+    .input-content,
+    .underline {
+      position: relative;
+    }
+
   </style>
 
   <template>
@@ -353,7 +358,7 @@ For example:
     },
 
     _computeInputContentClass: function(noLabelFloat, focused, _inputHasContent, _inputIsInvalid) {
-      var cls = 'input-content relative';
+      var cls = 'input-content';
       if (!noLabelFloat) {
         if (_inputHasContent) {
           cls += ' label-is-floating';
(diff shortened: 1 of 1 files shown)


notwaldorf:
@morethanreal yesterday's changes to layout broke `paper-input`, and this fixes it. PTAL.

bower.json
```diff
@@ -23,7 +23,9 @@
   "dependencies": {
     "polymer": "Polymer/polymer#v0.8.0-rc.7",
     "iron-input": "PolymerElements/iron-input#^0.8.0",
-    "paper-styles": "PolymerElements/paper-styles#^0.8.0"
+    "iron-autogrow-textarea": "PolymerElements/iron-autogrow-textarea#^0.8.0",
```
morethanreal: Where is this used?

notwaldorf: Argh, that's my checkout being wonky. Doesn't need to be here yet.

morethanreal:
Since `paper-styles` has a dependency on `iron-flex-layout`, shouldn't it get pulled in automatically?

notwaldorf:
Like this? The `relative` class gets added imperatively, so I thought it was easier to just make the class.

morethanreal:
I was thinking add these rules:

```css
.input-content {
    position: relative;
}
.underline {
    position: relative;
}
```

And in `_computeUnderlineClass` return `underline` instead of relative, and remove `relative` from value returned in `_computeInputContentClass`.

notwaldorf:
Done! :computer:

morethanreal:
lgtm

PR #22, opened 2015-05-07
Add a placeholder text

notwaldorf:
If the `paper-input` has a placeholder text and a label, then the label is always floating.


morethanreal:
Merged this code into `custom-validators`.

PR #23, opened 2015-05-07
support for custom validators and inputs

morethanreal:
Enables using a custom input in a paper-input-container. Either
provide attr-for-value to observe input value changes for validation
and label floating, or if unset the `input` event is used. See demo
for an example.

Also enables iron-validator-behavior support. The validator takes
precedence over constraints, if provided.

Bind to more HTMLInputElement attributes.

Renamed paper-autogrow-textarea to paper-textarea to be more
consistent with paper-textarea.

@cdata @notwaldorf PTAL

### DIFF
--- a/test/paper-input-error.html
+++ b/test/paper-input-error.html
@@ -31,7 +31,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
   <test-fixture id="auto-validate-numbers">
     <template>
-      <paper-input-container auto-validate>
+      <paper-input-container auto-validate attr-for-value="bind-value">
         <label id="l">label</label>
         <input is="iron-input" id="i" pattern="[0-9]*">
         <paper-input-error id="e">error</paper-input-error>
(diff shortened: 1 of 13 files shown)


test/paper-input-error.html
```diff
@@ -31,7 +31,7 @@
 
   <test-fixture id="auto-validate-numbers">
     <template>
-      <paper-input-container auto-validate>
+      <paper-input-container auto-validate attr-for-value="bind-value">
```
notwaldorf: is this needed? i saw that you removed it from `textarea` so i wasn't sure.

notwaldorf:
This is my favourite PR this year!

test/paper-input-error.html
morethanreal: Yeah, it's on purpose. The textarea is using the `input` event to update the value. It looks like `bind-value` didn't fire for some reason but have to debug in the other repository.

notwaldorf:
:sparkles:  LGTM :sparkles:

PR #26, opened 2015-05-13
allow manual validation

notwaldorf:
Added a `validate()` function to the behaviour, and refactored the `paper-input-container` to get notified (and care) when the contained `iron-input`'s validity changes. PTAL.

Needs https://github.com/PolymerElements/iron-input/pull/13 to work.

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -180,6 +180,13 @@ Use `Polymer.PaperInputBehavior` to implement your own `<paper-input>`.
       return this.$.input;
     },
 
+    /**
+     * Validates the input element and sets an error style if needed.
+     */
+     validate:function () {
+       return this.inputElement.validate();
+     },
+
     _computeAlwaysFloatLabel: function(alwaysFloatLabel, placeholder) {
       return placeholder || alwaysFloatLabel;
     }
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -311,7 +311,8 @@ For example:
     },
 
     listeners: {
-      'addon-attached': '_onAddonAttached'
+      'addon-attached': '_onAddonAttached',
+      'iron-input-validate': '_onInputValidityChanged'
     },
 
     get _valueChangedEvent() {
@@ -365,10 +366,13 @@ For example:
       var value = inputElement[this._propertyForValue] || inputElement.value;
 
       var valid;
-      if (inputElement.validate) {
-        valid = inputElement.validate(value);
-      } else if (inputElement.checkValidity) {
-        valid = inputElement.checkValidity();
+
+      if (this.autoValidate) {
+        if (inputElement.validate) {
+          valid = inputElement.validate(value);
+        } else {
+          valid = inputElement.checkValidity();
+        }
       } else {
         valid = true;
       }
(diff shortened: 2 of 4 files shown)


paper-input-behavior.html
```diff
@@ -180,6 +180,13 @@
       return this.$.input;
     },
 
+    /**
+     * Validates the input element and sets an error style if needed.
+     */
+     validate:function () {
+       return this.querySelector('paper-input-container').validate(this.inputElement, true);
```
morethanreal: It seems a little weird to me to have a `validate` function on `<paper-input-container>`, since it doesn't have validatable content but reflects the state of a validatable thing in its light DOM.

What do you think about calling `validate` on the input, which sets `invalid` (the property needs a `notify:true` in `iron-validatable-behavior`), and use binding to set the `invalid` property on the `<paper-input-container>`?

morethanreal: Or fire an event from `iron-validatable-behavior` that the container listens for to set `invalid`.

notwaldorf: Done!

paper-input-container.html
```diff
@@ -365,10 +366,14 @@
       var value = inputElement[this._propertyForValue] || inputElement.value;
 
       var valid;
-      if (inputElement.validate) {
-        valid = inputElement.validate(value);
-      } else if (inputElement.checkValidity) {
-        valid = inputElement.checkValidity();
+
+      // type="number" doesn't have a validate method...
```
morethanreal: Only `iron-input`'s have a `validate` method actually.

paper-input-container.html
```diff
@@ -365,10 +366,14 @@
...
+      if (this.autoValidate) {
+        if (inputElement.validate) {
+          valid = inputElement.validate(value);
+        } else {
+          valid = inputElement.checkValidity();
+        }
       } else {
         valid = true;
```
morethanreal: Does that work with `<input type="number">`? The check in line 382 needs the actual validity. Try typing in some alphabetic characters input a number input.

paper-input-container.html
```diff
@@ -311,7 +311,8 @@
     },
 
     listeners: {
-      'addon-attached': '_onAddonAttached'
+      'addon-attached': '_onAddonAttached',
+      'iron-input-validity-changed': '_onInputValidityChanged'
```
morethanreal: Maybe call the event `iron-input-validate`? Properties that are `notify:true` fires a  `*-changed` event, and might be confusing.

paper-input-container.html
notwaldorf: Not super sure, because `type="number"` is broken right now (because of https://github.com/PolymerElements/iron-input/issues/12), so typing anything breaks the element.

paper-input-container.html
notwaldorf: Done, here and in https://github.com/PolymerElements/iron-input/pull/13

paper-input-container.html
notwaldorf: Ok, I commented out the `selection` bit in `iron-input`, and it seems to work in the same way as it did before (it doesn't let you type letters)

morethanreal: Huh, how does that work\ without `auto-validate` and `prevent-invalid-input`?

notwaldorf: for `type="number"` it works by magic because the input itself doesn't let you type invalid input. For a regular input though, you're right, it sets it as invalid.

The problem is that I don't want to call `validate()` like we used to unless we need the value, since `validate()` now fires an event and will badge the decorator red if the input is invalid (which we don't need to do in the `auto-validate=false` case)

notwaldorf: Ah, I can fix this by adding a validation call to `iron-input`'s `_ensureValidValue`: https://github.com/PolymerElements/iron-input/pull/13

notwaldorf:
ping?

morethanreal:
Sorry, lgtm.

govis:
So if one calls .setCustomValidity('Test error') on the underlying input (is="iron-input") as per original 0.5 doc - should it automatically trigger paper-input's invalid state and display the error message? Doesn't seem to be the case.

<pre>
var pi = document.getElementById('test');
var i = Polymer.dom(pi.root).querySelector('input');
i.setCustomValidity('Test error');
</pre>

Issue #28, opened 2015-05-19
Invalid inputs look tragic in firefox


notwaldorf:
On `37.0.1`, they look like this:
![screen shot 2015-05-18 at 9 12 11 pm](https://cloud.githubusercontent.com/assets/1369170/7695312/b8cd4080-fda2-11e4-87cb-97f8ce0af2ad.png)

Text-areas and valid inputs look fine, though.

PR #29, opened 2015-05-19
fix native input's box-shadow

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/28

On Firefox, because CSS, an invalid `iron-input` has a box-shadow and shows up with a red box around it (see issue). 

After the fix, it looks like the paper-input we know and love
screen shot 2015-05-18 at 9 54 46 pm


morethanreal:
lgtm

PR #33, opened 2015-05-21
forward change events across the shadow boundary.

jklein24:
Let me know if this should go in the container or behavior instead.

### DIFF
--- a/paper-input.html
+++ b/paper-input.html
@@ -28,7 +28,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         <label>[[label]]</label>
       </template>
 
-      <input is="iron-input" id="input" bind-value="{{value}}" invalid="{{invalid}}" prevent-invalid-input="[[preventInvalidInput]]" type$="[[type]]" pattern$="[[pattern]]" maxlength$="[[maxlength]]" required$="[[required]]" disabled$="[[disabled]]" validator="[[validator]]" autocomplete$="[[autocomplete]]" autofocus$="[[autofocus]]" inputmode$="[[inputmode]]" minlength$="[[minlength]]" name$="[[name]]" placeholder$="[[placeholder]]" readonly$="[[readonly]]" size$="[[size]]">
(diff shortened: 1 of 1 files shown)


notwaldorf:
Hmm, why do you need this? The `iron-input`'s value is bound to the `paper-input`'s `value`, which has `notify:true` (see `PaperInputBehavior`). If that doesn't work, then it's a bug, but I don't think you should need to listen to the `iron-input` directly...

+ @morethanreal who knows all the things

orenb:
Listening to "on-input" works, but it fires an event on each character typed into the input. I would like to listen for the 'change' event which should only fire when the user is done modifying the input and unfocuses the input. These behaviors are slightly different.

morethanreal:
Can you add a test for the event firing on shadow DOM, and only a single event firing for shady DOM?

paper-input.html
```diff
@@ -54,8 +54,12 @@
 
     behaviors: [
       Polymer.PaperInputBehavior
-    ]
+    ],
 
+    /** Re-fire change event that does not bubble across shadow roots. */
+    _changeAction: function() {
+      this.fire('change');
```
morethanreal: Should probably only do this for native shadow DOM.

jklein24: Is Polymer.Settings.useShadow the right way to find that out?

morethanreal: @kevinpschaaf, is Polymer.Setting.useShadow ok for determining whether it's using native shadow DOM?

orenb:
Friendly ping: any updates on this?

kevinpschaaf:
You should use use `Polymer.Settings.useNativeShadow` to determine when to re-throw the event for now.

`Polymer.Settings.useShadow` will indicate when native SD or the SD polyfill is actually in use, and checking for `this.shadowRoot` is a better key since it means the element is actually _using_ a shadow root, but the SD polyfill does not appear to stop propagation of the `change` event per the spec (issue filed here: https://github.com/webcomponents/webcomponentsjs/issues/326), so until that bug is fixed `useNativeShadow` is what you'll need.

jklein24:
@kevinpschaaf, is there a good way to force shadowdom in a test? I guess I can fake it by setting useNativeShadow, but that seems sketchy.

kevinpschaaf:
@jklein24 You can either make a separate entry in the html file where the url has `?dom=shadow` on it, or else make a separate html file altogether with `<script>Polymer = { dom: 'shadow' };</script>` before `polymer.html`.

notwaldorf:
I'm going to close this PR, since we have a newer one to address this problem. Also check  https://github.com/PolymerElements/paper-input/pull/142/files#r37795543 for the solution we came up with

PR #38, opened 2015-05-22
add form behaviour

notwaldorf:
Since the behavior was removed in https://github.com/PolymerElements/iron-input/pull/23, it needs to live in the custom element now. PTAL.


morethanreal:
lgtm

PR #50, opened 2015-05-28
Fix an invalid param annotation

jklein24:


garlicnation:
LGTM only if the test failures aren't regressions.

PR #52, opened 2015-05-29
added info. on using with AngularJS

zishanj:
I took some time today to give it a test with AngularJS and it works perfectly with two-way data binding. So I thought to share it in readme file with others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,3 +3,12 @@
 `<paper-input>` is a Material Design text field.
 
 Contains a number of different features for validation, character counting, and more.
+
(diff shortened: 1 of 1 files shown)


zishanj:
I signed it!

notwaldorf:
Sorry for the late answer -- I'm desperately trying to catch up on PRs and issues. I'd like to not do this per element, but have a "Using Polymer Elements With Angular" guide on the docs site, where we do this generally. WDYT @arthurevans?

PR #70, opened 2015-06-08
add formatting helper

notwaldorf:
Added a helper formatting method (came up during https://github.com/PolymerElements/gold-phone-input/pull/16) that updates the value imperatively and preserves the caret position.

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -264,9 +264,31 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     /**
      * Validates the input element and sets an error style if needed.
      */
-     validate: function () {
+     validate: function() {
        return this.inputElement.validate();
      },
+
+    /**
+     * Restores the cursor to its original position after updating the value.
+     * @param {String} newValue The value that should be saved.
+     */
+    updateValueAndPreserveCaret: function(newValue) {
+      // Not all elements might have selection, and even if they have the
+      // right properties, accessing them might throw an exception (like for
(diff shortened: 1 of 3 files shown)


paper-input-behavior.html
```diff
@@ -264,10 +264,23 @@
     /**
      * Validates the input element and sets an error style if needed.
      */
-     validate: function () {
+     validate: function() {
        return this.inputElement.validate();
      },
 
+    /**
+     * Restores the cursor to its original position after updating the value.
```
morethanreal: doc for the param?

paper-input-behavior.html
```diff
@@ -264,10 +264,23 @@
...
+     */
+    updateValueAndPreserveCaret: function(newValue) {
+      var start = this.inputElement.selectionStart;
```
morethanreal: I wonder if you need to check for these properties, since somebody can use a custom element as the `inputElement`?

morethanreal:
Can we add a test for `<paper-textarea>` also?

paper-input-behavior.html
notwaldorf: Oh god, that, and I forgot (again) that `<input type=number>` can't use selectionStart.

paper-input-behavior.html
notwaldorf: :white_check_mark:

notwaldorf:
Added a textarea test and wrapped everything in a `try/catch`

morethanreal:
lgtm

PR #74, opened 2015-06-09
Fix tests on Windows IE/FF

morethanreal:
Simpler to assert on `classList` instead of `getComputedStyle` with fewer platform issues.

@notwaldorf PTAL


notwaldorf:
Nice! :+1: 

LOOK AT THOSE GREEN TESTS TOO :green_heart:

Issue #69, opened 2015-06-06
Applying rules from Polymer.PaperInputBehavior on textarea does not work


firasdib:
The `<paper-textarea>` element implements said behavior, but setting `always-float-label` has no effect.

morethanreal:
Fixed in https://github.com/PolymerElements/paper-input/pull/75

PR #75, opened 2015-06-09
Fix always-float-label

morethanreal:
For https://github.com/PolymerElements/paper-input/issues/69

@notwaldorf PTAL

### DIFF
--- a/paper-textarea.html
+++ b/paper-textarea.html
@@ -35,7 +35,7 @@ style this element.
 <dom-module id="paper-textarea">
   <template>
 
-    <paper-input-container no-label-float$="[[noLabelFloat]]" auto-validate$="[[autoValidate]]" disabled$="[[disabled]]" invalid="[[invalid]]">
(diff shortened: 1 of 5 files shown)


paper-textarea.html
```diff
@@ -35,7 +35,7 @@
 <dom-module id="paper-textarea">
   <template>
 
-    <paper-input-container no-label-float$="[[noLabelFloat]]" auto-validate$="[[autoValidate]]" disabled$="[[disabled]]" invalid="[[invalid]]">
+    <paper-input-container no-label-float$="[[noLabelFloat]]" always-float-label="[[_computeAlwaysFloatLabel(alwaysFloatLabel,placeholder)]]" auto-validate$="[[autoValidate]]" disabled$="[[disabled]]" invalid="[[invalid]]">
```
notwaldorf: Ugh, I wish there was a better way we didn't have to copy this in like 7 elements.

I'm just grumbling.

notwaldorf:
Should we add a test for `<input type="date">` that caused this? 

Either way, LGTM! :gem:

Issue #53, opened 2015-06-01
Label not floating in paper-input-container


tdesc:
Simple container with input and label. Expected behaviour that label will be float. It's not
Chrome desktop browser latest version.
```
    <paper-input-container>
        <label>Your name</label>
        <input is="iron-input">
    </paper-input-container>
```

notwaldorf:
The label should float once you start typing (but not initially). If you want it to float initially, you need to add the `always-float-label` attribute to the `paper-input-container`.

I've tried it both on Canary and Chrome stable, and it seems to work. Are you sure you're including all of the right things (`paper-input-container`, `iron-input`)?

tdesc:
paper-input works as expected. example as I wrote above dont.
Chrome MacOS version 43.0.2357.81 (64-bit)

notwaldorf:
Hmm, I'm not sure what you mean. Here is some code, in a jsbin: http://jsbin.com/cuqulajuto/3/edit.

The labels work fine in Chrome 43, Safari and Firefox. You're right that the char counter has a problem in Safari, and we're looking into that.

notwaldorf:
The remaining char-counter issue was fixed in https://github.com/PolymerElements/paper-input/pull/78

mixedCase:
I feel like I should add for the people googling this issue: if the label is not floating automatically it's very likely that you have not HTML imported iron-input.

PR #78, opened 2015-06-10
workaround unpredictable attached ordering

morethanreal:
Workaround until https://github.com/Polymer/polymer/issues/1773 is landed. For https://github.com/PolymerElements/paper-input/issues/53 PTAL @notwaldorf

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -305,10 +305,10 @@ This element is `display:block` by default, but you can set the `inline` attribu
       },
 
       _addons: {
-        type: Array,
-        value: function() {
-          return [];
-        }
+        type: Array
+        // do not set a default value here intentionally - it will be initialized lazily when a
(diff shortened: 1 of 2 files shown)


paper-input-container.html
```diff
@@ -369,6 +369,9 @@
     },
 
     ready: function() {
+      if (!this._addons) {
```
notwaldorf: Is this needed? (won't `_onAddonAttached` do it on the first addon anyway?)

morethanreal: Yeah, there could be no addons. I can lazily check for `this._addons` where it's referenced if you prefer.

notwaldorf: Oh, I was thinking that if there's no addons, we don't care if it's an array, but we probably assume it is somewhere else in the code. It's fine like this :)

notwaldorf:
lgtm

PR #80, opened 2015-06-11
Add a missing return in a default value function

jklein24:


morethanreal:
Nice, can you add a test case and also see why tests are failing? I think I fixed them such that they're all passing yesterday.

cdata:
Tests appear to be failing because CI Runner needs to be kicked ;)

jklein24:
Yeah the tests all pass locally for me.

morethanreal:
ugh, lgtm

PR #90, opened 2015-06-18
Typo fix - wrong wording for autoValidate property in doc

vuhrmeister:


vuhrmeister:
I signed it!

notwaldorf:
I think the original comment is correct. When setting the attribute in html, it has dashes, not camel case:
`<paper-input label="bork" auto-validate>`.

vuhrmeister:
Ok, it absolutely makes sense with `auto-validate` as the attribute. It was totally confusing to me with the documentation saying `autoValidate` as the property. But ok, that is the internal JS property, not the property name to set on the dom element.

I think I need to experiment a bit more to find out the proper usage of the new components since the docu seems to not be up-to-date at all.

notwaldorf:
@vuhrmeister Sorry about that. Here is more information about the attribute-property name mapping: https://www.polymer-project.org/1.0/docs/migration.html#property-name-to-attribute-name-mapping

Issue #85, opened 2015-06-11
"focused" property in paper-input-container missing notify:true


wq9:
I was wondering why `focused` was not working properly. Then I found out.

Focused property should have `notify: true` in `paper-input-container`
```
      /**
       * True if the input has focus.
       */
      focused: {
        readOnly: true,
        type: Boolean,
        value: false,
        notify: true
      },
```

PR #96, opened 2015-06-20
Add "notify: true" to focused property

wq9:
Focused property needs "notify: true" for upward one-way data binding.


wq9:
I signed it!

LM450N:
It will solves the issue #85 .

wq9:
Fixed with #147

PR #83, opened 2015-06-11
use IronControlState

morethanreal:
For https://github.com/PolymerElements/paper-input/issues/59 PTAL @cdata

Needs https://github.com/PolymerElements/iron-autogrow-textarea/pull/12 before merging.

PR #98, opened 2015-06-22
add back form behaviour

notwaldorf:
:wrench: the form behaviour that got accidentally :scissors: in https://github.com/PolymerElements/paper-input/pull/83


morethanreal:
lgtm

PR #104, opened 2015-06-29
add styling mixins for label and underline

### LINKED ISSUE #61: [1.0] Should have more complete set of custom properties
`<paper-input-container>` has `--paper-input-container-color` to control styling for the label and underline color together - but probably it should be possible to control label and underline separately simply by using custom properties?

morethanreal:
Resolves #61. PTAL @notwaldorf


notwaldorf:
Always :+1: custom properties!

PR #109, opened 2015-06-30
don't duplicate registration to the iron-form

notwaldorf:
Since the `iron-autogrow-textarea` now registers to be added to an `iron-form` (https://github.com/PolymerElements/iron-autogrow-textarea/pull/18), we don't need to also register in `paper-textarea`

PTAL @morethanreal


morethanreal:
lgtm

PR #112, opened 2015-07-07
use visibility instead of display for error

### LINKED ISSUE #95: Extra space created by auto validation error = BAD
https://github.com/Polymer/paper-input/issues/171

Elements above and below do not jump around to create extra space for the floating label = GOOD
image

This auto validation make me seasick because the elements below jump all over the place when you fill in a form = BAD
image

morethanreal:
Fixes #95. The error is no longer role="alert" which means
it does not announce, but it's ok for a11y because the input
still has invalid and aria-describedby.

PTAL @notwaldorf

### DIFF
--- a/paper-input-error.html
+++ b/paper-input-error.html
@@ -35,8 +35,8 @@ Custom property | Description | Default
   <style>
 
     :host {
-      /* need to use display: none for role="alert" */
-      display: none;
+      display: inline-block;
+      visibility: hidden;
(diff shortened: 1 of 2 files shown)


notwaldorf:
lgtm

morethanreal:
Looks like the tests are failing due to https://github.com/PolymerElements/iron-form-element-behavior/issues/8.

PR #115, opened 2015-07-09
add min, max, step

jscissr:
For type="number" or "range".
I tried to put them in alphabetically, but it's not really sorted.


jscissr:
I don't know how to properly set multiple types, so I simply set it to String. (There can be both numbers and dates as strings.)

notwaldorf:
This PR looks fine -- as a style preference, i would put the `min`/`max`/`step` props all together (maybe after minlength)?

jscissr:
I pushed a commit. I also put minlength and maxlength together.

notwaldorf:
Awesome, thanks so much!

PR #119, opened 2015-07-13
add prefix/suffix content selectors

notwaldorf:
New feature! This allows you to add a prefix/suffix element to the paper input (which in the long run should simplify all the duplication in the `gold-elements`). 

They look like this without input
screen shot 2015-07-13 at 3 05 41 pm

And like this with input:
screen shot 2015-07-13 at 4 23 13 pm

@morethanreal PTAL

P.S. All the tests are going to fail before https://github.com/PolymerElements/iron-form-element-behavior/pull/9 is merged :(

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -40,6 +40,8 @@
     "iron-validator-behavior": "PolymerElements/iron-validator-behavior#^1.0.0",
     "test-fixture": "PolymerElements/test-fixture#^1.0.0",
     "web-component-tester": "*",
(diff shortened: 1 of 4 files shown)


notwaldorf:
@morethanreal ping 🐳

morethanreal:
lgtm

morethanreal:
I lgtm'ed, but I'm starting to think that this element is doing way too much and maybe can be refactored.

notwaldorf:
@morethanreal As discussed, I'm going to land this and think about how we can refactor the whole thing, since that is a bigger adventure :(

PR #124, opened 2015-07-16
Adds rows and max-rows to textarea. #123.

### LINKED ISSUE #123: Add support for the rows and max-rows attribute to textarea.
Pass through `rows` and `max-rows` attributes to `iron-autogrow-textarea` so the paper-textarea can accept both `rows` and `max-rows`.

kevinwestern:
Resolves #123 .


notwaldorf:
@kevinwestern Thanks! The `rows` bit makes complete sense, but `max-rows` doesn't seem to make a difference? The textarea keeps growing nonetheless (and I can't find it in any html element docs...)

kevinwestern:
Hi! Yes, it's a bit confusing in that one would think `max-rows` wouldn't allow rows past that number but, taken from [iron-autogrow-textarea](https://elements.polymer-project.org/elements/iron-autogrow-textarea) it is:

> The maximum number of rows this element can grow to until it scrolls. 0 means no maximum.

notwaldorf:
Ohhhh that totally makes sense. Sorry! One more thing: could you add the comments from `iron-textarea` for `rows` and `maxRows` here as well, so they're picked up in the docs? Thanks!

PR #131, opened 2015-07-28
Adds rows and max-rows to textarea. Fixes #123.

### LINKED ISSUE #123: Add support for the rows and max-rows attribute to textarea.

kevinwestern:


notwaldorf:
Thanks!

PR #136, opened 2015-08-04
label should always stick to the underline

notwaldorf:
Otherwise suffixes/prefixes can push it up.

This doesn't affect normal inputs (sans any *-ixes), seeeeee:
screen shot 2015-08-03 at 5 13 55 pm

Note that with or without this patch, a giant label looks terrible and floats below the line. Future-Monica should fix this in a separate PR. With magical text truncation. Or something:
screen shot 2015-08-03 at 5 15 49 pm

/cc @morethanreal @cdata


cdata:
LGTM :+1:

PR #137, opened 2015-08-04
add prefix/suffix custom props

notwaldorf:
This allows styling the input and the -ixes (who may not necessarily be text) separately.

/cc @morethanreal @cdata


cdata:
LGTM :+1:

PR #140, opened 2015-08-10
fix #138

### LINKED ISSUE #138: Label misalignment issue
In some cases the label is being given an inline style like: 

```html
<div class="label-and-input-container style-scope paper-input-container">
  <label style="left: -26px;">...</label>
  <input is="iron-input">
</div>
```

That -26px in this case is unnecessary and causing misalignment of the label relative to the input. Seems to be coming in from https://github.com/PolymerElements/paper-input/blob/master/paper-input-container.html#L525

Repro case available internally at [go/paper-input-label-align-repro](http://go/paper-input-label-align-repro)

notwaldorf:
Fixes #138.

Needed to check if there were any `prefix` content children, not just whether the node existed (which it always does)

@cdata PTAL

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -521,7 +521,8 @@ This element is `display:block` by default, but you can set the `inline` attribu
           }
           // The label might have a horizontal offset if a prefix element exists
           // which needs to be undone when displayed as a floating label.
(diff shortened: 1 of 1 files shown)


cdata:
LGTM :+1:

PR #141, opened 2015-08-10
add disabled input mixin

### LINKED ISSUE #122: Add mixin for styling of disabled inputs
Sometimes you want to disable the input but still allow the user to read whats in the disabled input. Opacity of 0.33 is sometimes too low, so I suggest you allow the user to style this as they please with a mixin.

notwaldorf:
Fixes #122

@cdata PTAL


cdata:
LGTM :+1:

PR #143, opened 2015-08-11
Link to PaperInputContainer in docs

robdodson:


goldieep:
look good to me

abdonrd:
LGTM!

notwaldorf:
The behaviours are now linked in the catalog, so I don't think this is necessary (and no other element does this).

PR #147, opened 2015-08-17
make focused a notify property

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/85

@cdata PTAL


cdata:
Doesn't `focused` already notify?

https://github.com/PolymerElements/iron-behaviors/blob/master/iron-control-state.html#L26-L32

cdata:
Small question, otherwise LGTM :+1:

notwaldorf:
Different `focused` :sob: https://github.com/PolymerElements/paper-input/blob/master/paper-input-container.html#L350

PR #148, opened 2015-08-18
validate inputs on blur instead of attached

notwaldorf:
Validating on `attached` is a little weird, since it results in you looking at a form with all red fields you haven't really touched.

🎷@cdata

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -418,6 +418,10 @@ This element is `display:block` by default, but you can set the `inline` attribu
       return Polymer.dom(this).querySelector(this._inputSelector);
     },
 
+    get _inputElementValue() {
+      return this._inputElement[this._propertyForValue] || this._inputElement.value;
+    },
+
     ready: function() {
       if (!this._addons) {
         this._addons = [];
@@ -432,7 +436,12 @@ This element is `display:block` by default, but you can set the `inline` attribu
     },
 
     attached: function() {
-      this._handleValue(this._inputElement);
+      // Only validate when attached if the input already has a value.
+      if (this._inputElementValue != '') {
+        this._handleValueAndAutoValidate(this._inputElement);
+      } else {
+        this._handleValue(this._inputElement);
+      }
     },
 
     _onAddonAttached: function(event) {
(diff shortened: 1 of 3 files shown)


paper-input-container.html
```diff
@@ -453,20 +455,21 @@
...
+      this._handleValue(event.target, true);
     },
 
     _onValueChanged: function(event) {
-      this._handleValue(event.target);
+      this._handleValue(event.target, true);
     },
 
-    _handleValue: function(inputElement) {
+    _handleValue: function(inputElement, validateNow) {
```
samccone: I have fallen into this trap many times before
http://ariya.ofilabs.com/2011/08/hall-of-api-shame-boolean-trap.html

What would you think about something like 

```
_handleValue: function(inputElement, options) {
...

options.bypassValidation && +      if (validateNow && 
```

?

notwaldorf: Ideally, I wouldn't want any other parameters to `handleValue`, so maybe it's not as important now (but would be if this method ever gets changed again)

notwaldorf: Actually, I think I can do away with the entire boolean.

samccone: makes sense :)

paper-input-container.html
```diff
@@ -417,6 +417,10 @@
       return Polymer.dom(this).querySelector(this._inputSelector);
     },
 
+    get __inputElementValue() {
```
cdata: Private: a discerning choice. But! If you can't inherit from this thing (it is an element) isn't every protected method also a private method?

cdata: (btw this is :trollface: )

notwaldorf: Ughhhhhhh i'll make it `protected`. I'm clearly super confused about behaviours vs elements vs private things. I'll get there :rabbit:

paper-input-container.html
```diff
@@ -453,20 +458,21 @@
...
     },
 
     _onValueChanged: function(event) {
-      this._handleValue(event.target);
+      this._handleValue(event.target, true);
     },
 
-    _handleValue: function(inputElement) {
-      var value = inputElement[this._propertyForValue] || inputElement.value;
+    _handleValue: function(inputElement, validateNow) {
```
cdata: Once again, I'm not super excited about the boolean argument. It's kind of weird to read on lines 461, 465, and 469 for example. I'm not sure if there is a better way, though.

notwaldorf: The alternative is to split `handleValue` up and always call (whenever we passed `true`):
```
this._handleValue(event.target);
this._handleAutoValidate(event.target);
```

Do you think that's better? I'm ambivalent (copy pasting is bad vs booleans are bad)

cdata: Yes, I do like that better. At least it is obvious what is happening.

paper-input-container.html
cdata: Troll face means I'm trolling. It's fine for this to be private.

notwaldorf:
@cdata how about this? I wanted to avoid having to notify the add-ons twice :(

cdata:
Nice, this is looking cool :+1: :boat:

PR #157, opened 2015-08-24
Add focus passthrough for inputElement

### LINKED ISSUE #129: missing focus() method
Currently there is no way to programmatically focus a `paper-input`.

ktiedt:
Not sure if this is the best way to achieve this, I looked at the other behavior-* implementations for iron-input etc, and none seemed to encompass similar functionality except for ControlState maybe. So for now I included this in the paper-input-behaviors.

Fixes #129


notwaldorf:
@ktiedt I've already left comments in https://github.com/PolymerElements/paper-input/pull/162 before seeing this, so I'll close this one since it's the same PR essentially -- and it's not getting merged as-is anyway :)

PR #159, opened 2015-08-25
make invalid a notify property

notwaldorf:
:pager: @cdata


cdata:
This looks great :+1:

PR #162, opened 2015-08-28
Pass focus() to Input

kevashcraft:
Added a function to pass the focus command to the input element.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -65,6 +65,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       <button onclick="validate()">Validate!</button>
     </div>
 
+    <h4>Focus</h4>
+    <div class="vertical-section">
+      <paper-input label="textarea label" id="inputForFocus"></paper-input>
+      <br>
+      <button onclick="inputFocus()">Focus!</button>
+    </div>
+
     <h4>Character counter</h4>
     <div class="vertical-section">
       <paper-input label="label" char-counter></paper-input>
@@ -90,6 +97,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   <script>
     function validate() {
       document.getElementById('inputForValidation').validate();
+    }
+    function inputFocus () {
+      document.getElementById('inputForFocus').focus();
     }
   </script>
 </body>
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -237,6 +237,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       return this.$.input;
     },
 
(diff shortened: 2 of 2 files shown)


paper-input-behavior.html
```diff
@@ -237,6 +237,13 @@
       return this.$.input;
     },
 
+    /**
+     * Passes focus() to the input element
+     */
+    focus: function() {
+      this.$.input.focus();
```
notwaldorf: this should be `this.$.inputElement.focus()`

demo/index.html
```diff
@@ -65,6 +65,13 @@
       <button onclick="validate()">Validate!</button>
     </div>
 
+    <h4>Focus</h4>
+    <div class="vertical-section">
+      <paper-input label="textarea label" id="inputForFocus"></paper-input>
```
notwaldorf: I think `textarea` doesn't belong in this label :)

notwaldorf:
(+@cdata, our focus expert)

@kevashcraft I know this is a one line change, but it's actually really contentious. `focus` is a native method, which has a specific behaviour, which you're overriding. For example, calling `focus()` on an element without a tabindex doesn't actually focus that element. However, in this specific case, all `paper-input`s are inherently focusable (because their `iron-inputs` are), and setting a `tabindex`on a `paper-input` leads to a double-focus scenario (shift-tabbing out of a focused `paper-input` changes the tab from the `iron-input` to the `paper-input` container and looks weird).

I would be ok if you renamed the method to something like `focusInput`, and did the same thing (plus the comments I left on the code). I would also really like to see a test of this behaviour, not just in the demo (also, for the `paper-textarea`) 😊

Super sorry about all this. Focus is actually really hard to get right 😢

barakhub:
@notwaldorf, Sadly, Adding `focusInput` is not going to do the work.
Check out the following example: http://poly-play.appspot.com/C5FZY24iJmI
When the page opens, click on the 'Open modal' button and when the dialog opens hit shift+tab.

You'll see that the focus moves outside of the modal dialog to the 'Aleft 1' button. Hit space to see that it's really active.

The reason is here: https://github.com/PolymerElements/paper-dialog-behavior/blob/master/paper-dialog-behavior.html#L211
Once the focus moves outside of the dialog, the code in `PaperDialogBehavior._onFocus` tries to set it back to the last element with focus within the dialog, but if that's a `paper-input`, this fails (and this kind of change would fix that).

Generally, `paper-input` should "feel" like a built-in componnet and properly handle `focus`. Maybe it's better not to override `focus`, but instead handle the `focus` event.

notwaldorf:
Closing this PR, since the fix will be addressed by #186

PR #173, opened 2015-09-07
Replace translate3d by translateY

hypnoce:
Hello,

My team is building a mobile webapp using polymer. We managed to achieve very good performance by carefully measure and optimize bottlenecks. One of them is the number of graphical layers/GPU memory/Compositing. We found that the paper inputs have unnecessary layers due to the use of translate3d.

Thanks


notwaldorf:
Thanks! This makes sense! :)

Issue #178, opened 2015-09-14
Add type "search" to paper-input


Waxolunist:
Currently the type search is not allowed. Only number, password and text are allowed.
The type search would allow on iPhone to show on the keyboard instead of "Return" to display "Search".
Is it possible to add this behavior?

notwaldorf:
`type=search` seeeeeems to work, it just has a bit of a layout bug :)
```html
<paper-input label="label" type="search"></paper-input>
```
produces
<img width="298" alt="screen shot 2015-09-23 at 1 01 44 pm" src="https://cloud.githubusercontent.com/assets/1369170/10057075/462484d4-61f3-11e5-818e-4d07d80eea23.png">

I think the offset comes from the native html input which styles the "type=search" element differently :sob:
<img width="169" alt="screen shot 2015-09-23 at 1 06 12 pm" src="https://cloud.githubusercontent.com/assets/1369170/10057215/edec140c-61f3-11e5-8cc2-50f90444f3e3.png">

I'll take a look. Does everything else work fine for you, though?

notwaldorf:
I've also opened https://github.com/PolymerElements/paper-input/pull/182 to fix the styling a bit. The styling is still not great because browsers.

Waxolunist:
I added following to my css:

```

input[type=text]::-ms-clear {
  display: none;
  width: 0;
  height: 0;
}

input[type=text]::-ms-reveal {
  display: none;
  width: 0;
  height: 0;
}

input[type="search"]::-webkit-search-decoration,
input[type="search"]::-webkit-search-cancel-button,
input[type="search"]::-webkit-search-results-button,
input[type="search"]::-webkit-search-results-decoration {
  display: none;
}

input[type="search"] {
  -webkit-appearance: none;
}
```

To show the word search on the keyboard on iPhone, the input type search has to be surrounded with a form tag, which has an action attribute. So search is working actually.

You can close this one.

notwaldorf:
I've actually opened a PR to fix the styles, and add the results/auto save
attribute :)

PR #182, opened 2015-09-23
make type=search less terrible

notwaldorf:
As usual, `<input type=...>` is a jar of spiders. In today's re-enactment, `type=search` has hilarious and inconsistent styles across browsers. This PR makes it sort of work with `paper-input` (and sorta fixes https://github.com/PolymerElements/paper-input/issues/178). The end result is:

Chrome:
screen shot 2015-09-23 at 2 37 30 pm

FF: (look how nice it is. sigh)
screen shot 2015-09-23 at 2 38 16 pm

Safari:
screen shot 2015-09-23 at 2 38 27 pm
Safari also successfully displays the spyglass, but I haven't managed to trick the other 2 to do that:
screen shot 2015-09-23 at 2 49 33 pm

If you don't listen to the instructions and use a non-floated label, it looks amazing and there's nothing really that we can do about this (aside from putting in logic that forces you to auto-float the label if type="search". @cdata Should we do that? Sigh):
screen shot 2015-09-23 at 2 23 21 pm

I don't know what it looks on IE, but it's probably fine, right?

`<input>` or: why we can't have nice things. Tune in next week! :sob:

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -59,6 +59,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     <div class="vertical-section">
       <paper-input label="label"></paper-input>
 
+      <paper-input label="search" type="search" placeholder="type='search' should use placeholders instead of labels" autosave="test" results="5"></paper-input>
+
       <paper-input label="password" type="password"></paper-input>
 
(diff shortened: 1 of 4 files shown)


cdata:
Moar properties! LGTM :+1:

Issue #175, opened 2015-09-09
When an input has an error set, the text after it renders with an offset the length of the error


lkovaleva:
Code:
`<paper-input id="input"  label="Number input" pattern="\d+" auto-validate error-message="Please enter a number"></paper-input>`
`<div>Some label</div>`

Rendered:
![image](https://cloud.githubusercontent.com/assets/14207087/9775354/50cd9858-5703-11e5-9a8f-f807dd22e8d0.png)

The line right after the input is offset by the length of the error (when error is visible or not).

It looks like simply removing "float: left" from paper-input-error fixes this, but I'm not sure if that would cause any other adverse effect.

notwaldorf:
@lkovaleva As a quick fix, while I'm trying to figure out what's going on, adding this seems to fix it: `paper-input { display: inline-block; }`

PR #183, opened 2015-09-23
remove uneeded error float

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/175

(tldr: i don't know why that float is there, and taking it away makes everything look the same, minus borking anything that comes after it and is affected by the float)


cdata:
LGTM :+1:

notwaldorf:
Ugh, turns out removing this introduced some extra spacing. :(

PR #189, opened 2015-09-29
fix extra input-error spacing regression

notwaldorf:
This was introduced because of https://github.com/PolymerElements/paper-input/pull/183
Turns out the point of the `float` was to pull the next input up and not add extra spacing. 

Proof that it looks nice now:
screen shot 2015-09-29 at 2 19 06 pm

Floats: :skull:

### DIFF
--- a/paper-input-error.html
+++ b/paper-input-error.html
@@ -42,6 +42,7 @@ Custom property | Description | Default
 
         @apply(--paper-font-caption);
         @apply(--paper-input-error);
+        position: absolute;
       }
 
       :host([invalid]) {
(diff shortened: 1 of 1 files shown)


paper-input-error.html
```diff
@@ -50,7 +51,7 @@
     </style>
 
     <content></content>
-
```
cdata: Was this necessary for something?

cdata:
Tiny formatting issue, otherwise LGMT :+1:

paper-input-error.html
notwaldorf: the empty line? No, it's my editor being "tabs not spaces" or something. I didn't touch it.

Issue #167, opened 2015-09-01
Consider optimizing performance


sorvell:
Since users often create many instances of this element, it will especially benefit from performance optimization. Consider lazily creating optional or user initiated functionality (e.g. paper-ripple), leveraging behaviors over composition (e.g. paper-material), streamlining styling, and eliminating or simplifying features (e.g. behaviors)

esprehn:
The offsetParent in the template callback also causes a flood of style recalcs, one for every paper-input-container in the page:
https://github.com/PolymerElements/paper-input/blob/master/paper-input-container.html#L534

notwaldorf:
@esprehn Ah, that's my bad. This should only happen when `this.$.prefix` is true, but the bit that should bail out is last in the `if` statement. I'll move it to the top so that most of the time we don't even check if label has an `offsetParent`

esprehn:
@notwaldorf That still forces a synchronous recalc when the prefix is set, and also a synchronous distribution update since you call getDistributedNodes, polymer should really defer this work until requestAnimationFrame, and should batch the reads (calls to getDistributedNodes && offsetParent) and the writes of setting style.left

notwaldorf:
@sorvell should this still be open? this doesn't use `paper-ripple` nor `paper-material`, and I cleaned up the cray offset calcs

notwaldorf:
Closing, we've started working on https://github.com/PolymerElements/paper-text-field as a refactor, efficient version of this element

PR #190, opened 2015-09-29
less offsetParent recalc

notwaldorf:
This fixes the comment in https://github.com/PolymerElements/paper-input/issues/167#issuecomment-143912219

We don't really care about the `offsetParent` unless we have a prefix, so the `if` statement should bail out early if we don't.

I also think the `this.$.prefix` bit is always true because it just points to the `content` node, so I don't think it's needed.


cdata:
Nice :running_shirt_with_sash: LGTM :+1:

notwaldorf:
Would have been nicer if i wrote an if statement correctly the first time around :/

Issue #192, opened 2015-09-30
Long labels on small screens are sad


notwaldorf:
How sad, you ask? This sad:
![screen shot 2015-09-29 at 5 13 20 pm](https://cloud.githubusercontent.com/assets/1369170/10181477/81ea3f34-66cd-11e5-8a3d-403655b3b927.png)

Needs moar truncation or something

PR #193, opened 2015-09-30
fix long labels on small screens

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/192 with :sparkles: css magic :sparkles: 

Look at the nice truncated label:
screen shot 2015-09-29 at 5 20 15 pm

And we get extra space when we float:
screen shot 2015-09-29 at 5 50 19 pm

Unrelated: I've also deleted the `type=search` example from the demo, even though I just added it. It looks kind of weird, and we don't demo every other `type=`, so there's really no point (it does nothing special)

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -186,6 +186,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
         font: inherit;
         color: var(--paper-input-container-color, --secondary-text-color);
 
+        @apply(--paper-font-common-nowrap);
         @apply(--paper-font-subhead);
         @apply(--paper-input-container-label);
       }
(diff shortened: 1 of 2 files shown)


paper-input-container.html
```diff
@@ -186,6 +186,10 @@
         font: inherit;
         color: var(--paper-input-container-color, --secondary-text-color);
 
+        overflow: hidden;
+        white-space: nowrap;
+        text-overflow: ellipsis;
```
cdata: :gem:

notwaldorf:
@cdata there was already a mixin available in `paper-styles/typography.html', so I used that instead

Issue #180, opened 2015-09-18
--paper-input-container-prefix is documented but not implemented.


bendavis78:
There is no `@apply` for `--paper-input-container-prefix`.

bendavis78:
Actually, there appear to be three documented mixins that aren't implemented:

* --paper-input-container-input-disabled
* --paper-input-container-prefix
* --paper-input-container-suffix

In the stylesheet, they're implemented as:

* --paper-input-container-disabled
* --paper-input-prefix
* --paper-input-suffix

I'm not sure which of these are incorrect, the documentation or the code. My assumption is the latter, since all mixins in the docs use `--paper-input-container-*`.

notwaldorf:
I think the implementation is correct, at least for the suffix and the prefix (there's a `paper-input-error` mixin for the error as well, which doesn't have the container). Finally, `--paper-input-container-disabled` in particular is applied to the whole container, not just the input, so i think the documentation has a typo in that case as well.

TLDR: code is right, docs are wrong. Want to fire a PR for that, or should i? :)

bendavis78:
The convention I'm seeing is that all polymer elements define their own mixins using the element's name as a prefix. `--paper-input-error` is a mixin applied to the `:host` of the `paper-input-error` element (another convention is to use the element's name to mixin properties at that element's `:host` level). Since this is `paper-input-container`, and the prefix/suffix styles are applied within this element (and not in `paper-input`), I'm inclined to think the docs are correct.

notwaldorf:
Unfortunately I think that changing the current mixin names might break existing code that is using it, so I'd like to not do that. So even though it's not 100% kosher, I'd rather have the documentation be slightly inconsistent for the prefix/suffix mixins, than break existing code.

notwaldorf:
👉@cdata to double check he doesn't disagree

jacobevans:
I second changing the docs.
Sincerely,
A user who just wasted 20 minutes trying to get this to work with the wrong mixin name :smile:

PR #194, opened 2015-09-30
fix mixin docs

notwaldorf:
Updated the names of mixins incorrectly documented.
Fixes https://github.com/PolymerElements/paper-input/issues/180

As discussed in the issue, the suffix/prefix names are not consistent with the other `paper-input-container-*` names, but I think changing them might break existing code (where people have figured out what mixins to use correctly), which is not really worth it. So this is just making the docs not lie.


cdata:
LGTM :+1:

PR #195, opened 2015-10-01
Removed comma syntax error

ecoutu:
A comma in the middle of attribute declarations for
`<input is="iron-input" ...>` was causing html-minifier to spin in an
endless loop.


notwaldorf:
Ooof, thanks for catching this!

PR #201, opened 2015-10-10
Stop native validation kicking in when type is specified

MrAntix:
Check for a form associated input and the form[novalidate] attribute being set
if not found or the attribute not set then the input[type] is set to '' which stops native validation problems

a console message shows when this happens to give the developer a "heads up"
> type set to null as native validation is on, add novalidate attribute to the containing form

tests added

gold input will also have a change to set the input[type]="email", but I need to know the version of the dependency here before I do that one

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -344,7 +344,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
         this.type = '';
 
-        console.debug('type set to null as native validation is on, add novalidate attribute to the containing form');
+        if (console && console.debug) {
+          console.debug('type set to null as native validation is on, add novalidate attribute to the containing form');
+        }
       }
     },
 
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -342,7 +342,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       if (form == null
             || !form.hasAttribute('novalidate')) {
 
-        this.type = null;
+        this.type = '';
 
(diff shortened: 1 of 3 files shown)


notwaldorf:
I'm not sure I understand what is the issue that this PR is trying to solve. Can you put together a jsbin or link to the issue describing the problem?

MrAntix:
this came about from the gold email input not having an type set to 'email' meaning you don't get a nice email keyboard

https://github.com/PolymerElements/gold-email-input/pull/37

this was not set because it kicks in the native validation on a desktop

if you have form[novalidate] this stops the native validation

this change checks for the attribute and sets type=email on the gold input

I can do a jsbin if you really want, but will have to wait for the weekend

notwaldorf:
@MrAntix `gold-email-input` has had `type=email` for a while now (https://github.com/PolymerElements/gold-email-input/blob/master/gold-email-input.html#L65). Also, if this is just a fix for `gold-email-input`, it does not belong in `paper-input`, but in that element (which has its own `attached` method, so everything you do here can be done there

MrAntix:
ok, looks like some parallel dev going on

I was asked to move this from gold email to here
https://github.com/PolymerElements/gold-email-input/pull/37#issuecomment-142968777

happy its in anyway, I'll close this then

notwaldorf:
Ah, sorry about that. That was my bad. Basically, we ended up adding it to `gold-email` since it simplified a bazillion things, like the default regex validation. Sorry! :(

MrAntix:
not at all, glad its sorted :)

PR #191, opened 2015-09-29
Bind disabled attribute

kevashcraft:
Added an attribute to the iron-autogrow-textarea to pass the disabled state of paper-textarea.

Without this, the iron-autogrow-textarea is still focusable via javascript.

Note: needs [this pull request](https://github.com/PolymerElements/iron-autogrow-textarea/pull/44) to pass the attribute to the internal textarea

PR #205, opened 2015-10-14
bind disabled attribute

notwaldorf:
This just rebases and lands https://github.com/PolymerElements/paper-input/pull/191, which seems to have falled off the radar :)


cdata:
LGTM :+1:

PR #210, opened 2015-10-20
Master

oathcomrade:


paper-input-container.html
```diff
@@ -228,6 +228,7 @@
         box-shadow: none;
         padding: 0;
         width: 100%;
+        min-height: 26px;
```
notwaldorf: why is this needed?

paper-input-behavior.html
```diff
@@ -302,9 +302,27 @@
...
+    _checkForNativeFormatter: function() {
+      // only check these types if the label has to be always floated.
+      var checkTypes = ['number', 'time', 'date'];
+      if (checkTypes.indexOf(this.inputElement.type) === -1) {
+        return;
+      }
+      // TODO make value change silent
+      var value = this.inputElement.value;
+      // TODO catch chrome format warning
+      this.inputElement.value = ' ';
```
notwaldorf: Basically now every time you're attaching this input you're changing the value at least twice (once to empty, and once to the original value). Each of these has a bunch of side effects, and fires a bunch of observers, so it's a really bad approach when it comes to performance

notwaldorf:
This PR has no title, no description, and does not link to an issue, so I don't know what exactly it's trying to fix. Also, it's peppered with TODOs, which we tend to not allow.

notwaldorf:
Closing this since this looks inactive

Issue #106, opened 2015-06-30
paper-input: text-align


bwh-rl:
The iron-input in the paper-input element should inherit the text-align property.

With this small addition it wouldn't be necessary to use an own paper-input-container/iron input construct for a custom text-align.

Right now if the paper-input gets a text-align, the label changes, but not the input value.

PR #216, opened 2015-10-26
input inherits the text-align from the container

valdrinkoshi:
Fixes issue https://github.com/PolymerElements/paper-input/issues/106
Depends on https://github.com/PolymerElements/iron-autogrow-textarea/pull/51 when shadow dom is used.
Allows alignment of <code>paper-input, paper-textarea, paper-input-container</code> to be applied to the children (aka children will inherit the text-align from the container)
With this PR, users will be able to apply styles like:
```css
paper-textarea,
paper-input,
paper-input-container {
  text-align: center;
}
```

- takes care of placement of the label, input
screen shot 2015-10-26 at 10 54 24 am
- takes care of add-on-content
screen shot 2015-10-26 at 6 39 06 pm

Known issue: alignment of the label when paper-input has prefix/suffix content (to be tackled in a separate PR)
screen shot 2015-10-26 at 12 34 58 pm

### DIFF
--- a/paper-input-error.html
+++ b/paper-input-error.html
@@ -43,6 +43,8 @@ Custom property | Description | Default
         @apply(--paper-font-caption);
         @apply(--paper-input-error);
         position: absolute;
+        left:0;
+        right:0;
       }
 
       :host([invalid]) {
@@ -51,7 +53,7 @@ Custom property | Description | Default
     </style>
 
     <content></content>
-    
+
   </template>
 </dom-module>
 
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -237,6 +237,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
         border: none;
         color: var(--paper-input-container-input-color, --primary-text-color);
         -webkit-appearance: none;
+        text-align: inherit;
 
         @apply(--paper-font-subhead);
         @apply(--paper-input-container-input);
@@ -261,6 +262,10 @@ This element is `display:block` by default, but you can set the `inline` attribu
         resize: none;
       }
 
+      .add-on-content {
+        position: relative;
+      }
+      
       .add-on-content.is-invalid ::content * {
         color: var(--paper-input-container-invalid-color, --google-red-500);
       }


paper-input.html
```diff
@@ -67,6 +67,10 @@
         display: block;
       }
 
+      input {
```
notwaldorf: I think this actually belongs in `paper-input-container`. There's a bunch of other input styles there too.

Maybe we should do this for `textarea` as well?

valdrinkoshi: I was thinking of it, but then you'd have something like this:
screen shot 2015-10-26 at 11 23 12 am

is it the desired behavior?

notwaldorf: It shouldn't center the label, no. Just the text bits :)

valdrinkoshi: if we do
```css
paper-textarea {
  text-align: center;
}
```
we should get the label + textarea aligned on the center. Is this assumption right?

valdrinkoshi:
There are still some rendering issues:
- error labels wrongly placed (in this example it is not centered)
screen shot 2015-10-26 at 12 31 56 pm
- input label is not properly centered when input has icons on left/right
screen shot 2015-10-26 at 12 34 58 pm
- animation of the input label looks "weird", always flies from the right to the final position

Should we tackle these ones in a separate PR? or should we instead have a "text-align" property on the paper-input?

paper-input-error.html
```diff
@@ -43,6 +43,8 @@
         @apply(--paper-font-caption);
         @apply(--paper-input-error);
         position: absolute;
+        left:0;
+        right:0;
```
notwaldorf: I think this will make the error overlap `the char-counter` when it's too long:
screen shot 2015-10-26 at 3 49 34 pm

To be fair, I think this was happening before (it's happening without your patch), but it's definitely going to happen now.

valdrinkoshi: ok i'll fix this

notwaldorf: I'm an idiot -- I was running off the right branch, you actually fixed this a bit, and it's now not flying out of the container.

notwaldorf:
Thanks! LGTM :gem:

PR #218, opened 2015-10-26
Allow reusing the paper-input template

notwaldorf:
Moves the paper-input `<template>` to a named `dom-module`, so that `gold-*` elements can reuse it and stop copy pasting all of it.

Minor fixes:
- `PaperInputBehavior` already includes `IronControlState`, so don't include it twice
- should auto-validate on `ready()`. The `paper-input-container` already does this, but if the template
is cloned and we chose to manually do the auto-validation (i.e.

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -333,6 +333,12 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       this._updateAriaLabelledBy();
     },
 
+    ready: function() {
+      // If there's an initial input, validate it.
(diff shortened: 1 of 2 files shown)


MeTaNoV:
:+1: , working on a ```<gold-password-input>``` with an integrated strength meter and it is definitely tedious to mimic copy ```<paper-input>``` and bind the ```IronInputBehvior``` to the various element!

notwaldorf:
Going to close this PR for now and re-open it in a different state.

Issue #229, opened 2015-11-02
Floated label has incorrect size when using larger font-size for the field & label.


bendavis78:
[The MD spec for floated labels](https://www.google.com/design/spec/components/text-fields.html#text-fields-floating-labels) shows a [mock-up](http://i.imgur.com/CTcSj1B.png) with one of the fields having a larger font. The font-size of the field can be set using the `--paper-input-container-input` and `--paper-input-container-label` mixins. Unfortunately the floated label is scaled using `scale(0.75)`, which means it won't scale to the correct size if we override the font.

See [this jsbin](http://jsbin.com/fakete/21/edit?html,output) for a demo of the issue.

At the very least, there should be a mixin to allow custom styling of the label in the floating state 
(eg, `--paper-input-container-label-floating`).

PR #230, opened 2015-11-02
Added css mixin for floating label

bendavis78:
This is a workaround for #229.


notwaldorf:
Thanks! This is great. Could you also document this mixin in the `Styling` block in that file, please?

PR #236, opened 2015-11-05
add valueAsNumber property

### LINKED ISSUE #170: Support for valueAsNumber property
Like suggested in Polymer/polymer#1996, I am using an `iron-input`'s `valueAsNumber` property for two-way data binding to a value of type `Number` (called `theValue` in the example below) in my JSON model:

```html
<input is="iron-input" type="number" value-as-number="{{theValue::input}}">
```

I would like to replace this input with a `paper-input`. However, `paper-input` does not seem to support the `valueAsNumber` property. Could this be added? Or is there another way to bind a `paper-input`'s value to a number instead of a string?

valdrinkoshi:
Fixes #170 by adding valueAsNumber property
Depends on https://github.com/PolymerElements/iron-input/pull/58

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -64,6 +64,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       <paper-input no-label-float label="label (no-label-float)"></paper-input>
 
       <paper-input label="disabled" disabled></paper-input>
+
+      <paper-input label="with content" value="content"></paper-input>
+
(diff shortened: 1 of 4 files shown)


valdrinkoshi:
@notwaldorf @cdata FYI, please review

notwaldorf:
Going to close this PR for now, until we find a better approach.

Issue #228, opened 2015-10-30
Wrong floating-label placement with prefix usage


vinerz:
Hello everyone! I have a `readonly` `prefixed` and `prefilled` paper-input, with this markup:

```html
<paper-input name="username" id="username" label="Usuário" value="{{account.username}}" readonly always-float-label>
    <div prefix>http://example.com/</div>
</paper-input>
```

When the page renders, I get this result:
![Issue](https://sc-cdn.scaleengine.net/i/471716519dc857795ea792ce2d1ccf141.png)

Look at the floating-label misplaced after the prefix area!

When I give the paper-input focus, it fixes itself:
![Issue - Focused](https://sc-cdn.scaleengine.net/i/cb02560f9fe4d20669f9d7252a882ce3.png)

What can I do?

Thanks in advance,
Vinicius

notwaldorf:
What version of `paper-input` are you using? This should have been fixed in v1.0.18

vinerz:
Hello @notwaldorf! I've updated to the latest version 1.0.18, but the issue persists.

notwaldorf:
Ah, I can repro it. Thanks!

vinerz:
@notwaldorf, thanks for the answer, but the latest version of paper-input (1.1.0) still have this issue.

notwaldorf:
@vinerz Gah! I even have a test for this and I still missed an edge case. 😓

notwaldorf:
Oh, that's because the fix is in master and has not been released. It should be fixed in 1.1.1

vinerz:
@notwaldorf, just to confirm that this issue is fixed now. Thank you for your attention.

notwaldorf:
Whew! Good to hear :)

PR #243, opened 2015-11-09
fix prefix even moar better

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/228. Looking at the code some more, I think we don't need to listen to children at all, and just fix the floating label style to always align to the edge (which is a no-op if there's no prefix)

Bumped down the polymer dependency version as a result.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -27,7 +27,7 @@
   "homepage": "https://github.com/PolymerElements/paper-input",
   "ignore": [],
   "dependencies": {
-    "polymer": "Polymer/polymer#^1.2.0",
+    "polymer": "Polymer/polymer#^1.1.0",
     "iron-autogrow-textarea": "PolymerElements/iron-autogrow-textarea#^1.0.0",
     "iron-behaviors": "PolymerElements/iron-behaviors#^1.0.0",
     "iron-form-element-behavior": "PolymerElements/iron-form-element-behavior#^1.0.0",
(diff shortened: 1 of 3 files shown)


bower.json
```diff
@@ -27,7 +27,7 @@
   "homepage": "https://github.com/PolymerElements/paper-input",
   "ignore": [],
   "dependencies": {
-    "polymer": "Polymer/polymer#^1.2.0",
```
cdata: It's a little weird that a previous version of this element requires v1.2.0, but a more recent version requires v1.1.0.. but, I guess it's not any weirder than the rest of our dependency graph.

paper-input-container.html
```diff
@@ -562,16 +547,15 @@
 
         if (alwaysFloatLabel || _inputHasContent) {
           cls += ' label-is-floating';
+          // If the label is floating, ignore any offsets that may have been
+          // applied from a prefix element.
+          this.$.labelAndInputContainer.style.position = 'static';
```
cdata: :star:

cdata:
I assume there are already tests that cover this positioning?

Otherwise, LGTM :+1:

bower.json
notwaldorf: I don't have to change it. I think I was being overly pedantic, maybe. Should I revert it?

notwaldorf:
@cdata there are no tests that cover positioning for this label :(
I'll look into it.

bower.json
cdata: No, it's fine. It's just weird!!

cdata:
Thanks for tests :100: LGTM :+1:

PR #248, opened 2015-11-12
stop bringing in the paper-styles kitchen sink

notwaldorf:
Removes any use of the `/deep/` flex classes. One step closer away from dragons 🐉🎉 

Before:
screen shot 2015-11-12 at 3 31 37 pm

After:
screen shot 2015-11-12 at 3 33 02 pm

And hopefully nothing else :(

(I would release this as a minor version bump, P.S)


cdata:
LGTM :+1:

PR #258, opened 2015-11-22
add the reflectToAttribute to invalid property

MeTaNoV:
Hi,
while developing a custom element based on `<paper-input-container>`, I needed an add-on that required to know the invalid state of the component. I therefore, decided to use:
```css
      :host-context([invalid]) {
        visibility: hidden;
      };
```
It does not work however, because `invalid` is not reflected to attributes.
Looking at `<paper-input-error>`, I discovered that an invalid property is duplicated and itself reflected to attributes.
This PR simplifies and reduces the code accordingly.
I also updated the relevant test.

### DIFF
--- a/paper-input-error.html
+++ b/paper-input-error.html
@@ -48,7 +48,7 @@ Custom property | Description | Default
         right:0;
       }
 
-      :host([invalid]) {
+      :host-context([invalid]) {
         visibility: visible;
       };
     </style>
@@ -65,20 +65,5 @@ Custom property | Description | Default
     behaviors: [
       Polymer.PaperInputAddonBehavior
     ],
-
-    properties: {
-      /**
-       * True if the error is showing.
-       */
-      invalid: {
-        readOnly: true,
-        reflectToAttribute: true,
-        type: Boolean
-      }
-    },
-
-    update: function(state) {
-      this._setInvalid(state.invalid);
-    }
   });
 </script>
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -361,6 +361,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
       invalid: {
         observer: '_invalidChanged',
         type: Boolean,
+        reflectToAttribute: true,
         value: false
       },
 
--- a/test/paper-input-error.html
+++ b/test/paper-input-error.html
@@ -54,6 +54,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         var error = Polymer.dom(container).querySelector('#e');
         assert.equal(getComputedStyle(error).visibility, 'hidden', 'error is visibility:hidden');
         input.bindValue = 'foobar';
+        assert(container.invalid, 'invalid should be set to true');
+        assert(container.hasAttribute('invalid'), 'invalid attribute should be set');
         assert.notEqual(getComputedStyle(error).visibility, 'hidden', 'error is not visibility:hidden');
       });
 


MeTaNoV:
@cdata kindly PTAL! :smile_cat:

cdata:
Thanks for the contribution!

I personally like this usage, but there are performance consequences to reflecting to attribute. @notwaldorf is currently refactoring this element to improve performance, so she should be the one to approve this change.

MeTaNoV:
The attribute is reflected in all case in the addon, it is just a factorisation optimisation, since I am forced also to add it and reflect it in my own addon.

notwaldorf:
@MeTaNoV I guess the bigger question is: what do you want to style in the invalid case? There are already a bunch of mixins that are applied to different parts of the input when the validity changes, and those aren't styled using `[invalid]`, so I don't think you could stomp over those styles anyway.

MeTaNoV:
Like I said, `<paper-input-error>` is introducing an `invalid ` property (which is also reflected to attributes) in the unique purpose to style itself properly. I would have to do it also for my addon (which I temporally did). Instead, I propose to have it on the parent, so that we could reduce from a tiny amount the code due to factorization ... Nothing less, nothing more! 
Thanks for the follow up!

MeTaNoV:
@notwaldorf next one? :dancer:

paper-input-error.html
```diff
@@ -65,20 +65,5 @@
     behaviors: [
       Polymer.PaperInputAddonBehavior
     ],
-
-    properties: {
-      /**
-       * True if the error is showing.
-       */
-      invalid: {
```
notwaldorf: I don't think this can be removed, since it would constitute a breaking change, no? What if someone is using this attribute?

MeTaNoV: Nobody is supposed to use this one, since its presence is only relevant to style himself. The invalidity is semantically a property of its container.

notwaldorf: I mean it's a public property on an element that you can use. That's the definition of a breaking change

MeTaNoV: Well, from this perspective yes. The question is, why did this property was introduced? I think the answer is to style itself only, which is IMHO unoptimized. As the owner of the component, you are the one to take responsibility for the optimization change I proposed. I mean, a lot of PRs won't never be accepted, I wouldn't mind don't worry! :)

notwaldorf: I think this can be be cleaned up in a 2.0 release, but maybe we shouldn't merge this for now. Adding the appropriate label for tracking

MeTaNoV: sure! and HNY BTW! :)

jfrazzano: If one binds to invalid$=[[item.inputInvalid]] , when loading an array  of objects to be spit through a dom-repeat to create a variety of inputs for a form, an empty value causes a styling reversion to a native input. 

True false and no bind do not have the effect.  The effect is not replicable outside the Dom repeat. 

Not sure if anyone else had the issue. 

If it's intentional it allows for a non-standard usage: typing the paper input to radio radio button and checkbox and using suffix plus no label float to crest check box and radio sets 

I don't mind the option, but do not believe it is intended.

notwaldorf:
Closing this PR, as we can't merge it right now (since it's a breaking change), and this element is undergoing a fair bit of refactoring in a parallel repo (`paper-text-field`), in which `paper-input-error` is not a standalone element

PR #259, opened 2015-11-25
Polymer.IronControlState duplicate inheritance

MeTaNoV:
Polymer.IronControlState is already inherited in Polymer.PaperInputBehavior


notwaldorf:
Thanks!

PR #266, opened 2015-12-03
center the input-related-elements vertically

notwaldorf:
This way the prefixes, sufixes and the label will be aligned vertically (vs. aligned at the bottom, which makes them look sad)

/cc @ebidel


ebidel:
LGTM!

cdata:
LGTM :+1:

PR #268, opened 2015-12-04
Remove unused import and fix rel attribute

abdonrd:
- Remove paper-input-char-counter.html import because is unused
- Fix rel attribute from color.html


notwaldorf:
I think this needs a rebase...

abdonrd:
Rebased!

PR #269, opened 2015-12-04
Fix #229 - css mixin for floating label with docs

### LINKED ISSUE #229: Floated label has incorrect size when using larger font-size for the field & label.

chiefcll:
Closes #230

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -81,6 +81,7 @@ Custom property | Description | Default
 `--paper-input-container-disabled` | Mixin applied to the container when it's disabled | `{}`
 `--paper-input-container-label` | Mixin applied to the label | `{}`
 `--paper-input-container-label-focus` | Mixin applied to the label when the input is focused | `{}`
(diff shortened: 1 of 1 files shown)


chiefcll:
I signed it!

notwaldorf:
Thanks!

PR #270, opened 2015-12-05
Sort imports alphabetically and remove/add some spaces

abdonrd:


paper-input-container.html
```diff
@@ -177,9 +176,10 @@
       }
 
       .input-content {
-        position: relative;
         @apply(--layout-horizontal);
         @apply(--layout-center);
+
+        position: relative;
```
notwaldorf: why is this change needed?

paper-input-container.html
```diff
@@ -198,18 +198,17 @@
 
       .input-content.label-is-floating ::content label,
       .input-content.label-is-floating ::content .paper-input-label {
+        /* Since we scale to 75/100 of the size, we actually have 100/75 of the
+        original space now available */
+        width: 133%;
```
notwaldorf: why is this change needed?

paper-input-container.html
abdonrd: It is not necessary, but most of the elements have the external `@apply()`'s at the beginning.

I think it would be awesome to have a style guide of styles for the elements. :smiley:

paper-input-container.html
abdonrd: I dont't know why this change. :confused:

Back!

paper-input-container.html
notwaldorf: Yeah a style guide would be neat. I think this particular change is ok, but moving CSS around just for the sake of moving it makes me a bit worried, since sometimes before or after a mixin has a meaning (`--layout-fit`, for example, applies `position:relative`, so in some cases if you want to override that, before or after the mixin is applied is actually important)

abdonrd: Yes, you're absolutely right! **Cascading** Style Sheets... :grin:

Issue #275, opened 2015-12-10
Sizing issues in ie


robrez:
If a paper-input is used inside of a horizontal flex container, it's size is teeny-tiny...

eg:
```
<div class="layout horizontal justified">
  <paper-input label="one"></paper-input>
  <paper-input label="two"></paper-input>
</div>
```

The issue can be seen via this jsbin --
http://jsbin.com/necala/edit?html,output

The same general issue has been documented in a couple of other repos --
here...
https://github.com/PolymerElements/iron-flex-layout/pull/9
and here...
https://github.com/PolymerElements/paper-tabs/issues/29

notwaldorf:
Fixes by #276

PR #276, opened 2015-12-10
Use flex-auto to fix IE

robrez:
Fix for issue ..
https://github.com/PolymerElements/paper-input/issues/275

paper-input

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -172,7 +172,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
       }
 
       .label-and-input-container {
-        @apply(--layout-flex);
+        @apply(--layout-flex-auto);
         @apply(--layout-relative);
       }
 


notwaldorf:
Thanks!

notwaldorf:
@robrez Argh, unfortunately this causes problems with `<paper-textarea>`, which now acts funky when given a small width, like:
```css
paper-textarea {
  width: 30px;
}
```

robrez:
Oh man! So sorry about that!

##### --paper-input-container-input to the rescue?
I'm thinking, instead of trying to hammer every nail, I could just make use of the `--paper-input-container-input` mixin. I'll try and retest this theory and update my findings.

##### Here are some details about another case that made me worry about the changes from this PR...

This past week I was trying to "fix" gold-cc-input for IE using the flex-basis trick. The overall markup was the same as this case -- gold-cc-input inside a horizontal flex container.
- w/o flex-auto, element was sized way too small
- w/ flex-auto, element was sized correctly
- unless element was explicitly sized (like your case -- width: 200px).
-- When the element was explicitly sized, the paper-input's contents overflow the paper-input's boundry

I was trying to figure some solution to fix all of the problematic-in-ie cases known to me, but I got discouraged and never got back around to it.
Here are some additional changes I was testing out. They seemed to achieve the desired results but I don't feel confident about them at all.

```
      .label-and-input-container {
        @apply(--layout-flex-auto);
        @apply(--layout-relative);
+       width: 100%
+       max-width: 100%
      }
...
      .input-content ::content input,
      .input-content ::content textarea,
      .input-content ::content iron-autogrow-textarea,
      .input-content ::content .paper-input-input {
        position: relative; /* to make a stacking context */
        outline: none;
        box-shadow: none;
        padding: 0;
        width: 100%;
+       max-width: 100%;
        background: transparent;
        border: none;
        color: var(--paper-input-container-input-color, --primary-text-color);
        -webkit-appearance: none;
        text-align: inherit;
        @apply(--paper-font-subhead);
        @apply(--paper-input-container-input);
      }
```

I'm worried that the changes above might also have some adverse effects on `prefix` and/or `suffix`.

These IE-only problems make me :*(
[flexbugs](https://github.com/philipwalton/flexbugs) has been a useful resource

robrez:
Tinkering around today, I added a new mixin to this rule in paper-input-container--

```
      .label-and-input-container {
        @apply(--layout-flex);
        @apply(--layout-relative);
+       @apply(--paper-input-blah);
      }
```

Then elsewhere..

```
    .hdiv {
        @apply(--layout-horizontal);
        @apply(--layout-justified);
        @apply(--layout-wrap);
    }

    paper-input.ie-fix {
      --paper-input-blah: {
        @apply(--layout-flex-auto);
      };
    }
...
    <div class="hdiv">
        <paper-input class="ie-fix" label="text input"></paper-input>
        <paper-input class="ie-fix" label="text input"></paper-input>
    </div>
```

Confirmed this works for the original case.

The other thought I had was making a custom-property for 'width'.
Then a rule in paper-input-container could do something like...
```
      .label-and-input-container {
        @apply(--layout-flex);
        @apply(--layout-relative);
      }

      .label-and-input-container.oh-hai-width {
        flex-basis: var(--paper-input-width);
      }
```

The closest I've come to a rule that makes everything seem to work OK is this one --
```
      .label-and-input-container {
-       @apply(--layout-flex);
+       @apply(--layout-flex-auto);
        @apply(--layout-relative);
+       width: 100%
+       max-width: 100%
      }
```

¯\\_(ツ)_/¯

notwaldorf:
I don't think adding a `--paper-input-width` custom property is a good idea. It's too specific, and the `width` property on the host should already has the same intent.

I think the last thing you said, so adding 
```
.label-and-input-container {
        @apply(--layout-flex-auto);
        @apply(--layout-relative);
        width: 100%;
        max-width: 100%;
      }
```

might make the most sense, and seems to work well. Would you mind sending a PR for that, referencing this PR in the description, please? Thanks!

robrez:
What do you think about these additional changes --

Either:
```

      ::content [prefix] {
        @apply(--paper-font-subhead);
        @apply(--paper-input-prefix);
+       @apply(--layout-flex-none);
      }

      ::content [suffix] {
        @apply(--paper-font-subhead);
        @apply(--paper-input-suffix);
+       @apply(--layout-flex-none);
      }
```
or
```

      ::content [prefix] {
        @apply(--paper-font-subhead);
        @apply(--paper-input-prefix);
+       -ms-flex-shrink: 0;
+       -webkit-flex-shrink: 0;
+       flex-shrink: 0;
      }

      ::content [suffix] {
        @apply(--paper-font-subhead);
        @apply(--paper-input-suffix);
        @apply(--layout-flex-none);
+       -ms-flex-shrink: 0;
+       -webkit-flex-shrink: 0;
+       flex-shrink: 0;
      }
```

Without them, prefix and suffix shrink -- see the top example below.

prefixsuffix

robrez:
Sorry for spamming you --


Also, retesting paper-text area.
Before this PR (and after), placing two paper-textarea elements inside of a horizontal flex parent causes the incorrect "small sizing" problem, even in chrome.

Given this markup..
```
    <demo-snippet class="horizontal">
      <template>
        <paper-textarea label="autoresizing textarea input"></paper-textarea>
        <paper-textarea label="autoresizing textarea input"></paper-textarea>
      </template>
    </demo-snippet>
```

textarea-nosize

With or w/o the previous PR, and even the intended follow-up PR, paper-textarea requires some explicit sizing rule whenever included in a horizontal flex parent..
textarea-sized

I don't think this is something I want to try to fix --
adding a `--flex-auto` to the host rule of `paper-textarea` will give the element size, but prevents the element containing the paper-textarea from setting an explicit size rule eg `width: 50px`.

notwaldorf:
The flex container bug is separate (#299), so you don't have to worry about that. Also, your original PR didn't break that, so it's an unrelated issue that I'll fix. For now, I just want to fix this one breakage :)

I would also hold off the prefix/suffix changes. Maybe send them as a separate PR/issue, since I'm not sure what exactly the problem is (i.e. when do they shrink?)

robrez:
I think they shrink whenever they're iron-icons, I was leaning towards fixing them here because the new rule `width: 100%; max-width: 100%` causes the problem

notwaldorf:
Ohhhh I see. Ok, let's do those too in the same PR, and continue the conversation there (so I can pull the code and test it for a bit. I'm losing track of what the changes are 😟)

PR #293, opened 2015-12-30
bind title attribute

jab:
This PR attempts binds the title attribute that a paper-input is invoked with to the title attribute of the native input element in its template. This is useful because when you provide a pattern attribute that is not matched by the user's input, the browser will show your title to the user to help them figure out how to correct their input.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

> Use the title attribute to describe the pattern to help the user.

See also "Title Attribute" section of http://www.html5rocks.com/en/tutorials/forms/constraintvalidation/

### DIFF
--- a/paper-input.html
+++ b/paper-input.html
@@ -100,6 +100,7 @@ style this element.
         aria-labelledby$="[[_ariaLabelledBy]]"
         aria-describedby$="[[_ariaDescribedBy]]"
         disabled$="[[disabled]]"
+        title$="[[title]]"
         bind-value="{{value}}"
         invalid="{{invalid}}"
         prevent-invalid-input="[[preventInvalidInput]]"


jab:
Here is a screenshot showing off Chrome using the title set on an invalid paper-input for the validity message of the native input when the user tries to submit the containing form:
screen shot 2016-01-02 at 10 15 41

jab:
Meant to say, this change improves Polymer interoperability with Webshims, a popular library enabling html5 form validation in browsers that don't yet implement it (including latest Safari for iOS and OS X), and recommended at https://github.com/Modernizr/Modernizr/wiki/HTML5-Cross-Browser-Polyfills#web-forms.

Without this change, if you set a `pattern` on a `paper-input` and Webshims' form validation polyfill is active, it will actually log an error to the console:
screen shot 2016-01-03 at 11 24 33

paper-input-behavior.html
```diff
@@ -170,6 +170,13 @@
       // HTMLInputElement attributes for binding if needed
 
       /**
+       * Bind this to the `<input is="iron-input">`'s `title` property.
+       */
```
notwaldorf: I don't think `title` needs to be defined as a property, since it's a native attribute (and therefore always present)

paper-input.html
```diff
@@ -90,7 +90,7 @@
       }
     </style>
 
-    <paper-input-container no-label-float="[[noLabelFloat]]" always-float-label="[[_computeAlwaysFloatLabel(alwaysFloatLabel,placeholder)]]" auto-validate$="[[autoValidate]]" disabled$="[[disabled]]" invalid="[[invalid]]">
+    <paper-input-container no-label-float="[[noLabelFloat]]" always-float-label="[[_computeAlwaysFloatLabel(alwaysFloatLabel,placeholder)]]" auto-validate$="[[autoValidate]]" disabled$="[[disabled]]" invalid="[[invalid]]" title="[[title]]">
```
notwaldorf: I don't think `paper-input-container` needs the `title` attribute, since it's not doing anything with it

paper-input.html
```diff
@@ -100,6 +100,7 @@
         aria-labelledby$="[[_ariaLabelledBy]]"
         aria-describedby$="[[_ariaDescribedBy]]"
         disabled$="[[disabled]]"
+        title="[[title]]"
```
notwaldorf: I think it should be `title$=`

jab:
Sorry for the n00b mistakes and thanks for reviewing, @notwaldorf! Fixed in ddefce1 and force-pushed. I also correspondingly fixed and force-pushed on #296, so if that one looks good, merging #296 should automatically close this PR at the same time.

Issue #277, opened 2015-12-10
paper-input temporarily creates a horizontal scrollbar after typing first letter


johnbradley:
This can be seen in the demo.
https://elements.polymer-project.org/elements/paper-input?view=demo:demo/index.html&active=paper-input
Just type a letter in one of the paper-input textboxes where the label is setup to float upwards.
While the label is floating upwards there is a horizontal scrollbar.
It only occurs when the first character is typed since that is when the label floats upward.

It looks like it is related to the "width: 133%;" specified in paper-input-container.html.
```
      .input-content.label-is-floating ::content label,
      .input-content.label-is-floating ::content .paper-input-label {
      ....
          /* Since we scale to 75/100 of the size, we actually have 100/75 of the
          original space now available */
          width: 133%; 
```
Changing this to width:100% seems to fix the issue, but probably messes up whatever is going on with the scaling.

Gwalks91:
+1

MeTaNoV:
I have the same strange behavior in one of my demo

robrez:
I worked around it like this:
```
      paper-input,
      paper-textarea {
        position: relative;
        overflow: hidden;
      }
```

notwaldorf:
Fixed in #294

PR #294, opened 2015-12-31
paper-input temporarily creates a horizontal scrollbar after typing first letter #277

johnbradley:
I added "overflow: hidden" to paper-input and paper-textarea to prevent the scrollbar.
This is a variant of robrez's fix in #277.
Not sure of the larger implications of this change.

### DIFF
--- a/paper-input.html
+++ b/paper-input.html
@@ -71,6 +71,7 @@ style this element.
     <style>
       :host {
         display: block;
+        overflow: hidden;
       }
 
       input::-webkit-input-placeholder {
--- a/paper-textarea.html
+++ b/paper-textarea.html
@@ -37,6 +37,7 @@ style this element.
     <style>
       :host {
         display: block;
+        overflow: hidden;
       }
     </style>
 


notwaldorf:
I've tested this for a bit and it doesn't seem to have any unintended side effects (famous last words)

robrez:
Hate to be the bearer of bad news...

This truncates `error-message`...
err

##### I think, maybe, adding the overflow rule to this rule in paper-input-container, instead of paper-input and paper-textarea might do the trick...
```
      .input-content {
        @apply(--layout-horizontal);
        @apply(--layout-center);

        position: relative;
        overflow: hidden;
      }
```
It may fix other custom usages of paper-input-container, as well, if any of them had the original problem (eg: gold-cc-input, etc).


Alternatively -- 
I noticed that char-counter isn't truncated.
I think this is because `paper-input-error` uses absolute positioning; whereas, char-counter is floated.

Changing paper-input-error to be floated (left) instead of absolute positioned prevents truncation.. but it also causes the paper-input to be taller than paper-input that has no `error-message` property. Changing paper-input slightly can make the added height only kick-in when the element is invalid..

```
      <template is="dom-if" if="[[_showErrorMessage(errorMessage, invalid)]]">
        <paper-input-error>[[errorMessage]]</paper-input-error>
      </template>
...
    _showErrorMessage: function(errorMessage, invalid) {
      return invalid && (errorMessage && errorMessage.length > 0);
    }
```

But I think the reasoning behind the absolute positioning is because it would be undesirable for the form-height to jump around as errors come and go

notwaldorf:
:sob: I am tempted to revert this PR, in that case.

notwaldorf:
@robrez adding that rule to `input-content` makes the label float incorrectly (dissapear completely)

:sob: I'm tempted to revert this PR since the breakage >> what it was fixing

johnbradley:
I say revert this PR. It certainly made things worse.

johnbradley:
Worked with @robrez trying to come up with a solution.
This is what we came up with:
The problem is the transformation of the label in paper-input-container.html.
```
.input-content.label-is-floating ::content label,
        -webkit-transform: translateY(-75%) scale(0.75);
        transform: translateY(-75%) scale(0.75);
        -webkit-transition: -webkit-transform 0.25s;
        transition: transform 0.25s;
        -webkit-transform-origin: left top;
        transform-origin: left top;
        /* Since we scale to 75/100 of the size, we actually have 100/75 of the
        original space now available */
        width: 133%;
```

We transform the scaling via:
```        -webkit-transition: -webkit-transform 0.25s;
        transition: transform 0.25s;```
But the width is applied immediately.
So the label stretches out to 133% making the scrollbar.
Then over 0.25 seconds the scaling takes place.
Once both are applied we aren't getting too big.
The solution that seems to work is adding width to the transformation.
It looks like this will only work if we have already applied width earlier.
I will send a pull request soon.

johnbradley:
Here is the pull request: #304.
This undoes the previous pull requests changes as well.

PR #296, opened 2016-01-02
bind oninvalid

jab:
This enables binding to a paper-input's contained native input's `oninvalid` attribute, which allows setting completely custom error messages with something like `<paper-input ... oninvalid="setCustomValidity('Custom Message')")>`:

screen shot 2016-01-02 at 11 13 40

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -224,6 +224,13 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       },
 
       /**
+       * Bind this to the `<input is="iron-input">`'s `oninvalid` property.
+       */
+      oninvalid: {
+        type: String
+      },
+
+      /**
        * Limits the numeric or date-time increments.
        * Bind this to the `<input is="iron-input">`'s `step` property.
        */
--- a/paper-input.html
+++ b/paper-input.html
@@ -100,6 +100,7 @@ style this element.
         aria-labelledby$="[[_ariaLabelledBy]]"
         aria-describedby$="[[_ariaDescribedBy]]"
         disabled$="[[disabled]]"
+        title$="[[title]]"
         bind-value="{{value}}"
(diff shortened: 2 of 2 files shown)


jab:
This is a follow-on to #293 and was branched off that branch, so if this is merged it should merge and close that PR at the same time.

paper-input-behavior.html
```diff
@@ -224,6 +224,13 @@
...
+       * Bind this to the `<input is="iron-input">`'s `oninvalid` property.
+       */
+      oninvalid: {
+        type: String
+      },
+
+      /**
        * Limits the numeric or date-time increments.
        * Bind this to the `<input is="iron-input">`'s `step` property.
        */
```
notwaldorf: `onInvalid` is not a property, it's an event listener for the native `input's` `invalid` event: https://developer.mozilla.org/en-US/docs/Web/Events/invalid. It's not fired every time the validity is checked, but whenever `checkValidity` is called.

Here's a Jsbin explaining this: http://jsbin.com/yonoho/edit?html,console,output

I think that if you want to listen to the `invalid` event, it's fair that you'd have to listen on the nested element (so `paperInput.inputElement`), since it's a non-bubbling event. However, maybe it would be better if `iron-input` called `checkValidity` in its validate() method, rather than just checking the `valid` property.

jab: Thanks @notwaldorf and sorry for more n00b mistakes! Is PolymerElements/iron-input#70 what you meant?

notwaldorf: Yesss, but let me test that for a bit since it's a fairly big change. In any case, I think this PR can be closed :)

jab: Sounds good, closing. Thanks @notwaldorf.

notwaldorf: No worries! Glad I could help. Looking at the other PR now :)

Issue #288, opened 2015-12-20
polylint error


opus1269:
```
$ polylint paper-input.html
paper-input.html:99:7
    Property 'tabindex' bound to attribute 'tabindex$' not found in 'properties' for element 'paper-input'
```

notwaldorf:
I think this may be a polylint issue. `tabindex` is a native property available on all elements. /cc @garlicnation

notwaldorf:
Closing this, as I've opened it on the `polylint` repo

PR #300, opened 2016-01-04
Satisfying polylint

sandro-k:
Polylint is giving the following output:

```
$ polylint --input demo/index.html 
paper-input.html:99:7
    Property 'tabindex' bound to attribute 'tabindex$' not found in 'properties' for element 'paper-input'
```

After adding the `tabindex` property to the `paper-input-behavior` the error is gone. If you want the `tabindex` to be before the private `_ariaDescribedBy` and `_ariaLabelledBy` property I can update my PR.

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -316,8 +316,14 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       _ariaLabelledBy: {
         type: String,
         value: ''
-      }
+      },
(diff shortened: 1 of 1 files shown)


notwaldorf:
As mentioned in https://github.com/PolymerElements/paper-input/issues/288, this is a `polylint` bug. All PolymerElements are HTMLElements, which inherently have a `tabindex` property

PR #304, opened 2016-01-06
paper-input temporarily creates a horizontal scrollbar after typing first letter #277 #294

johnbradley:
fixing temporary scrollbar again
Removing the overflow hidden that killed the error labels.
Changing width to animate with scaling so the strange width:133% + scale(0.75) stuff will be applied at the same rate. Had to apply an initial width to get this to work.
I tested on windows in chrome, IE and Edge.
Didn't notice any issues.

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -189,6 +189,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
         top: 0;
         right: 0;
         left: 0;
+        width: 100%;
         font: inherit;
         color: var(--paper-input-container-color, --secondary-text-color);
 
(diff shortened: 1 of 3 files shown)


paper-input-container.html
```diff
@@ -201,8 +202,8 @@
       .input-content.label-is-floating ::content .paper-input-label {
         -webkit-transform: translateY(-75%) scale(0.75);
         transform: translateY(-75%) scale(0.75);
-        -webkit-transition: -webkit-transform 0.25s;
-        transition: transform 0.25s;
+        -webkit-transition: -webkit-transform 0.25s, width 0.26s;
```
notwaldorf: Would you mind adding a comment, please, as to why this is needed? Just so future us aren't weirded out by the 0.01s difference?

paper-input-container.html
```diff
@@ -189,6 +189,7 @@
         top: 0;
         right: 0;
         left: 0;
+        width: 100%;
```
notwaldorf: is this needed?

notwaldorf:
@johnbradley Thanks for digging through this! i pulled it locally and tested floating labels/rtl/error messages/prefixes and it looks great!

paper-input-container.html
johnbradley: Firefox seems to need this or it doesn't animate the width and we get the temporary scroll bar.
I think if it doesn't have a value already set it doesn't animate the width property.
Chrome doesn't need this.

paper-input-container.html
johnbradley: The idea was to make sure the scaling was finished before the width was finished.
I don't think it was necessary so I made them the same. In my mind the scaling and width will occur incrementally together and be OK. Didn't notice any issues retesting after this change.

notwaldorf:
Thanks!

PR #306, opened 2016-01-06
Autofloat label for special types

### LINKED ISSUE #291: paper-input with type="date" issue
I try to use paper-input with type="date" and i got a problem with this:
`<paper-input ngDefaultControl type="date" label="birthday" ></paper-input>`
typedate
and some issue like this
datepinput

notwaldorf:
Fixes #291 .
Despite saying in the docs you shouldn't do this, it's a bit surprising when the label overlaps some native input's default placeholder (things like `type="date"` always have a placeholder). So I'm admitting defeat and fixing this, and auto-floating the label in that case.

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -394,8 +394,20 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       return this.inputElement;
     },
 
+    registered: function() {
+      // These types have some default placeholder text; overlapping
+      // the label on top of it looks terrible. Auto-float the label in this case.
+      this._typesThatHaveText = ["date", "datetime", "datetime-local", "month",
+          "time", "week", "file"];
+    },
+
     attached: function() {
       this._updateAriaLabelledBy();
+
+      if (this.inputElement &&
+          this._typesThatHaveText.indexOf(this.inputElement.type) !== -1) {
+        this.alwaysFloatLabel = true;
+      }
     },
 
     _appendStringWithSpace: function(str, more) {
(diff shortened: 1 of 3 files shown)


paper-input-behavior.html
```diff
@@ -356,6 +356,15 @@
 
     attached: function() {
       this._updateAriaLabelledBy();
+
+      // These types have some default placeholder text; overlapping
+      // the label on top of it looks terrible. Auto-float the label in this case.
+      var typesThatHaveText = ["date", "datetime", "datetime-local", "month",
+          "search", "file", "time", "week"];
```
cdata: This seems worth storing in a static so that we don't create an array for every time every input is attached. :sake:

paper-input-behavior.html
```diff
@@ -356,6 +356,15 @@
 
     attached: function() {
       this._updateAriaLabelledBy();
+
+      // These types have some default placeholder text; overlapping
```
cdata: Did you check browser consistency of this detail?

cdata:
Why are the tests :crying_cat_face: ?

notwaldorf:
Oh hmm I had fixed the tests but I guess didn't push correctly. Yay git.

paper-input-behavior.html
notwaldorf: All of the date ones only work in Chrome, otherwise browsers default to `text`. The `file` inputs are a little different and inconsistent, but the `placeholder` attribute is always available to be used (and looks like an unfloated label), so I'd rather err on the "always float in this case" rather than try to split it on browsers. WDYT?

paper-input-behavior.html
notwaldorf: Did I do it right? I put it in `registered`

cdata:
LGTM :+1:

Issue #260, opened 2015-11-27
Unclear text in docs


zoechi:
Several attributes have this text in the documentation 

> Bind this to the <input is="iron-input">'s xxxxx property.

I have no clue what this means. Do I have to to something, how? Is this what the element does itself?

notwaldorf:
That is the documentation for the `PaperInputBehavior`, which is a behaviour that you could use to implement your own paper-input like thing. In the docs they show up because they're used by `paper-input`. You don't have to bind them, they're already bound for you by the element. I will try to make the docs clearer in this case

PR #307, opened 2016-01-06
add more deets to the behaviour docs

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/260

Since the behaviour docs appear in the `<paper-input>` page, there's some confusion about what "bind this to the iron-input" actually means (since you don't actually have an iron-input, as an user). Cleared up the docs to explain this only matters if you're using the behaviour to roll out your own element.

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -32,7 +32,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        */
 
       /**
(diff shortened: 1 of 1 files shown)


cdata:
LGTM :+1: Nice docs :scroll: 

Looks like IE10 is :trollface:

notwaldorf:
@cdata Yeah, I'll look into the tests in a different PR. They look borked on master (and on every other input PR i have)

PR #308, opened 2016-01-12
Automatic cleanup!

azakus:


cdata:
This is a transitive failure due to PolymerLabs/polylint#90 which incorrectly flags paper-input as doing something bad.

notwaldorf:
Closing this PR, as I think @azakus sent a new one to the same effect. Also this is old and needs a rebase.

PR #311, opened 2016-01-14
Update to the new paper theme colors

abdonrd:
Also, sort apply()s

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -74,7 +74,7 @@ The following custom properties and mixins are available for styling:
 Custom property | Description | Default
 ----------------|-------------|----------
 `--paper-input-container-color` | Label and underline color when the input is not focused | `--secondary-text-color`
-`--paper-input-container-focus-color` | Label and underline color when the input is focused | `--default-primary-color`
+`--paper-input-container-focus-color` | Label and underline color when the input is focused | `--primary-color`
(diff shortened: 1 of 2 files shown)


paper-input-container.html
```diff
@@ -260,18 +264,19 @@
         -webkit-appearance: none;
         text-align: inherit;
 
-        @apply(--paper-font-subhead);
```
notwaldorf: Why does this need to be moved?

paper-input-container.html
```diff
@@ -195,8 +199,6 @@
         font: inherit;
         color: var(--paper-input-container-color, --secondary-text-color);
 
-        @apply(--paper-font-common-nowrap);
```
notwaldorf: why do these need to move? :(

paper-input-container.html
abdonrd: Like the other Polymer elements, if we can (because is cascading), set the external `apply()`s at the top. I turn back?

notwaldorf: See answer in https://github.com/PolymerElements/paper-dialog-behavior/pull/61/files#r49791795

abdonrd:
@notwaldorf updated! :sparkles:

notwaldorf:
Thank you so much! 🙏

PR #328, opened 2016-02-07
Apply the transition effect for floating labels moving back to their …

Pesthuf:
…non-floated state.

paper-input-containers with floating labels (that don't have always-float-label set) move their label into the top right corner and shrink it down when text is entered into the input.
There's a short transition that animates this transition.

However, when deleting the text so that the input ends up empty, the label snaps back to its initial position in the input immediately, with no transition applied. This happens because the transition is only defined in CSS for when the input-content has the class **label-is-floating**.

This pull request applies the transition for the label in both states, floating and not floating, so that when it goes from its floating state back to its initial non-floating state, it will still be animated.

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -196,6 +196,10 @@ This element is `display:block` by default, but you can set the `inline` attribu
         width: 100%;
         font: inherit;
         color: var(--paper-input-container-color, --secondary-text-color);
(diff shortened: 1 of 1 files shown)


Pesthuf:
I signed the CLA.

notwaldorf:
Neat, thanks so much for the fix! The label looks much shinier now ✨

Issue #321, opened 2016-01-29
Old focus code in paper-input-behaviour needs to be removed


notwaldorf:
There's some hacky looking code in PaperInputBehaviour(https://github.com/PolymerElements/paper-input/blob/master/paper-input-behavior.html#L500) that worked around focus problems in IronControlState that do not exist, and I am suspicious they're causing weird focus behaviour (like https://github.com/PolymerElements/paper-input/issues/318). Removing it shouldn't break anything, since we handle focus and blur separately.

PR #331, opened 2016-02-09
clean up focus/blur tests

notwaldorf:
- fixes https://github.com/PolymerElements/paper-input/issues/321
- uses the overriden `_focusBlurHandler` from `IronControlState` instead of redefining a `focus` handler
- hopefully cleans up the tests so that they make more sense, and add a giant comment about what the current state of focus/blur is
- removes a test from `paper-textarea` because it was from before the time when textarea didn't have a tabindex by default

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -365,12 +365,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     listeners: {
       'addon-attached': '_onAddonAttached',
-      'focus': '_onFocus'
     },
-
-    observers: [
-      '_focusedControlStateChanged(focused)'
-    ],
 
     keyBindings: {
       'shift+tab:keydown': '_onShiftTabDown'
@@ -440,12 +435,17 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     /**
-     * Forward focus to inputElement
+     * Forward focus to inputElement. Overriden from IronControlState.
      */
-    _onFocus: function() {
-      if (!this._shiftTabPressed) {
+    _focusBlurHandler: function(event) {
+      if (this._shiftTabPressed)
+        return;
+
+      Polymer.IronControlState._focusBlurHandler.call(this, event);
+
+      // Forward the focus to the nested input.
+      if (this.focused)
         this._focusableElement.focus();
-      }
     },
 
     /**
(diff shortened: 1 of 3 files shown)


paper-input-behavior.html
```diff
@@ -365,13 +365,8 @@
 
     listeners: {
       'addon-attached': '_onAddonAttached',
-      'focus': '_onFocus'
     },
 
-    observers: [
-      '_focusedControlStateChanged(focused)'
-    ],
```
cdata: :+1:

paper-input-behavior.html
```diff
@@ -440,12 +435,17 @@
     },
 
     /**
-     * Forward focus to inputElement
+     * Forward focus to inputElement. Overriden from IronControlState.
```
cdata: :beers:

paper-input-behavior.html
```diff
@@ -440,12 +435,17 @@
...
-    _onFocus: function() {
-      if (!this._shiftTabPressed) {
+    _focusBlurHandler: function(event) {
+      if (this._shiftTabPressed)
+        return;
+
+      Polymer.IronControlState._focusBlurHandler.call(this, event);
+
+      // Forward the focus to the nested input.
+      if (event.target === this && event.type === 'focus')
```
cdata: Is it sufficient to reduce this conditional to:

```javascript
if (this.focused) {
  // ...
}
```

:question:

notwaldorf: Looks like it! Done.

test/paper-input.html
```diff
@@ -196,23 +196,51 @@
...
       });
 
-      test('focus/blur events fired on host element', function() {
-        var nFocusEvents = 0;
-        var nBlurEvents = 0;
+      // At the moment, it is very hard to correctly fire exactly
+      // one focus/blur events on a paper-input. This is because
+      // when a paper-input is focused, it needs to focus
+      // its underlying native input, which will also fire a `blur`
+      // event.
```
cdata: While this may be true, there should be some expectation that a `blur` event is fired at some point, right?

notwaldorf: Hmm, what do you mean? `blur` will also be fired when you blur the `paper-input`, obviously. But it will also fire when you focus the `paper-input`, which is I think unexpected (and what the comment was trying to explain)

cdata: Previously, this test described a particular behavior on the host: that it fires focus events when it is focused, and blur events when it is blurred. Now, the test only asserts the former.

The expectation that it fires a set number of times is no longer practical, but my feedback is that it still makes sense to set the expectation that `blur` will fire.

notwaldorf: Oh! I just split them up into 2 different tests! There's the equivalent `blur events fired on host element` tests below.

cdata: Ahh yes, sorry I have been distracted by :fire: today. SGTM.

test/paper-textarea.html
```diff
@@ -139,33 +133,50 @@
...
+
+      test('blur events fired on host element nested element is blurred', function() {
+        MockInteractions.focus(input);
+        input.addEventListener('blur', function(event) {
+          assert(!input.focused, 'input is blurred');
+        });
+        MockInteractions.blur(input.inputElement.textarea);
+      });
+
+      test('focusing then bluring sets the focused attribute correctly', function() {
```
cdata: :100:

cdata:
LGTM :+1:

PR #342, opened 2016-02-18
add IronFormElementBehavior to paper-textarea

notwaldorf:
Fixes https://github.com/PolymerElements/iron-autogrow-textarea/issues/33

This is part of a 3-headed hydra of PRs that is trying to fix the following problem: `iron-form` was bad at de-duping nested elements with the same `name`, and would double submit them. The problem was that if you bind the `name` attribute all the way down, and these children you're binding it to are either native elements (so the form automatically cares about them) or custom elements with the `IronFormElementBehavior`, you would submit the same value twice.

This PR adds the correct `IronFormElementBehavior` to the element, since it's an element that wants to work in a form, and that's what the behavior does.

Landing this without the other 3 means that a `paper-textarea` will be submitted twice.

Related PRs:
- https://github.com/PolymerElements/iron-autogrow-textarea/pull/74
- https://github.com/PolymerElements/iron-form/pull/110

### DIFF
--- a/paper-textarea.html
+++ b/paper-textarea.html
@@ -10,6 +10,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
 <link rel="import" href="../polymer/polymer.html">
 <link rel="import" href="../iron-autogrow-textarea/iron-autogrow-textarea.html">
+<link rel="import" href="../iron-form-element-behavior/iron-form-element-behavior.html">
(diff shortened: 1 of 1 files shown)


valdrinkoshi:
LGTM

Probably related to `bind-value` cleanup, but you could update the internal `iron-autogrow-textarea` to use `value` instead of `bind-value`
```html
<iron-autogrow-textarea id="input" class="paper-input-input"
        value="{{value}}"

PR #348, opened 2016-02-26
Fix SSN input layout in demo

### LINKED ISSUE #347: Custom inputs inside container breaks
The `paper-input-container` forces the width of `input` to 100% causing demo to break.

http://polygit2.appspot.com/components/paper-input/demo/index.html

selection_010

srikkbhat:
fixes #347

### DIFF
--- a/demo/ssn-input.html
+++ b/demo/ssn-input.html
@@ -9,6 +9,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 -->
 <link rel="import" href="../../polymer/polymer.html">
 <link rel="import" href="../../iron-input/iron-input.html">
+<link rel="import" href="../../iron-flex-layout/iron-flex-layout.html">
 <link rel="import" href="ssn-validator.html">
 
 <dom-module id="ssn-input">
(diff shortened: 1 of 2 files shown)


demo/ssn-input.html
```diff
@@ -27,15 +28,23 @@
...
     <ssn-validator></ssn-validator>
 
-    <input is="iron-input" maxlength="3" bind-value="{{_ssn1}}" size="3" aria-label="First 3 digits of social security number">
-    -
-    <input is="iron-input" maxlength="2" bind-value="{{_ssn2}}" size="2" aria-label="Middle 2 digits of social security number">
-    -
-    <input is="iron-input" maxlength="4" bind-value="{{_ssn3}}" size="4" aria-label="Last 4 digits of social security number">
+    <div class="container">
+
+      <input is="iron-input" maxlength="3" bind-value="{{_ssn1}}" size="3" aria-label="First 3 digits of social security number">
```
notwaldorf: I think you can just set a style for  `input { width: auto; }` to undo the `100%`, and don't need `iron-flex-layout`

srikkbhat:
The input width is `auto` in the current code.  The style from `paper-input-container` takes precedence over the style provided here.
https://github.com/PolymerElements/paper-input/blob/v1.1.8/demo/ssn-input.html#L27

I used the solution from `gold-cc-expiration-input` which has similar scenario and uses width 100% along with iron-flex-layout.
https://github.com/PolymerElements/gold-cc-expiration-input/blob/v1.1.1/date-input.html#L45

PR #349, opened 2016-02-29
<paper-textarea> placeholder color styles correctly

hotforfeature:
Adds in styling for `<paper-textarea>`'s placeholder to match `<paper-input>`. You can see the color difference on the current release with the following snippet:

```html
<paper-input label="Label" placeholder="I'm Right!"></paper-input>
<paper-textarea label="Label" placeholder="I'm sad :("><paper-textarea>
```

It's even more obvious if you set either a `--paper-input-container-color` or `--secondary-text-color`.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -29,7 +29,7 @@
   "ignore": [],
   "dependencies": {
     "polymer": "Polymer/polymer#^1.2.0",
-    "iron-autogrow-textarea": "PolymerElements/iron-autogrow-textarea#^1.0.0",
+    "iron-autogrow-textarea": "PolymerElements/iron-autogrow-textarea#^1.0.11",
     "iron-behaviors": "PolymerElements/iron-behaviors#^1.0.0",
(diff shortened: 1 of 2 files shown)


bower.json
```diff
@@ -29,7 +29,7 @@
   "ignore": [],
   "dependencies": {
     "polymer": "Polymer/polymer#^1.2.0",
-    "iron-autogrow-textarea": "PolymerElements/iron-autogrow-textarea#^1.0.0",
+    "iron-autogrow-textarea": "PolymerElements/iron-autogrow-textarea#^1.0.11",
```
notwaldorf: This change isn't needed -- ^1.0.0 uses the latest version, and your change doesn't break if you have a lower version than `1.0.11`. Since you're just styling the placeholder, it would be a no-op if you had a lower version, so it's still fine

notwaldorf:
Going to close this PR until the comment is addressed. Please re-open when it's ready for another look!

PR #351, opened 2016-03-02
Fix closure compiler error (using Chromium settings).

tommycli:
Hello,

This fixes a closure compiler issue when we compile this with the Chromium settings.

Note, that this header is duplicated text from PaperInputAddonBehavior. I tried using @override, but that did not work. I'm open to suggestions if there's a better way to do this.

If accepted, can we also roll the version so Chromium can pull the new version of paper-input? Thanks.

Tommy

### DIFF
--- a/paper-input-char-counter.html
+++ b/paper-input-char-counter.html
@@ -65,6 +65,17 @@ Custom property | Description | Default
       }
     },
 
+    /**
+     * This overrides the update function in PaperInputAddonBehavior.
+     * @param {{
+     *   inputElement: (Element|undefined),
+     *   value: (string|undefined),
+     *   invalid: boolean
+     * }} state -
+     *     inputElement: The input element.
+     *     value: The input value.
+     *     invalid: True if the input value is invalid.
+     */
     update: function(state) {
       if (!state.inputElement) {
         return;
@@ -72,7 +83,7 @@ Custom property | Description | Default
 
       state.value = state.value || '';
 
-      var counter = state.value.length;
+      var counter = state.value.length.toString();
 
       if (state.inputElement.hasAttribute('maxlength')) {
         counter += '/' + state.inputElement.getAttribute('maxlength');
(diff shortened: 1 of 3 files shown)


mgiuffrida:
@notwaldorf any chance we could get this and a version bump in today?

paper-input-char-counter.html
```diff
@@ -72,7 +83,7 @@
 
       state.value = state.value || '';
 
-      var counter = state.value.length;
+      var counter = state.value.length.toString();
```
notwaldorf: Why is this `toString` needed?

notwaldorf:
LGTM with one comment about the `toString`

paper-input-char-counter.html
mgiuffrida: because it's being used as a string and assigned to a string-typed member?

addyosmani:
LGTM2. I'm unfortunately none the wiser on the need for the additional `toString()` other than it being more explicit for that string-typed member.

tommycli:
Yes exactly, there's a closure compiler error without the explicit toString. Thanks!!

mgiuffrida:
Yes, the Closure compiler compares the `@type`s of declared variables and will error when the wrong type is assigned.

```
(ERROR) Error in: paper-input-char-counter-extracted.js
## /usr/local/google/home/michaelpg/dev/c/tot/src/third_party/polymer/v1_0/components-chromium/paper-input/paper-input-char-counter-extracted.js:39: ERROR - assignment to property _charCounterStr of PaperInputCharCounterElement
## found   : (number|string)
## required: string
##       this._charCounterStr = counter;
##       ^
```

See, `counter` can be either a `Number` or a `String`:

```
         var counter = state.value.length;
  
         if (state.inputElement.hasAttribute('maxlength')) {
           counter += '/' + state.inputElement.getAttribute('maxlength');
         }
  
         this._charCounterStr = counter;
```

but `_charCounterStr` is only declared as a `String`.

Issue #313, opened 2016-01-14
Dynamically creating polymer element that expect a child causes error.


justinfagnani:
Report from internal customer:

When dynamically creating a 'paper-input-container' element:
    var elt = document.createElement('paper-input-container');
I get the following error because it expects to have an 'input' child:
    Uncaught TypeError: Cannot read property 'addEventListener' of undefined

This is caused by the ready function:
  https://github.com/PolymerElements/paper-input/blob/master/paper-input-container.html#L458

KrauseStefan:
I have the same issue when using angular 2, where content is construct dynamically one element at the time.

The problem is very simular to https://github.com/PolymerElements/paper-dialog-scrollable/issues/22

I've made a temporary fix based on this, but I still have some issues so I'm not aiming for a pull request.
https://github.com/KrauseStefan/paper-input/tree/Feature-support-dynamical-creation

PR #354, opened 2016-03-08
set up children listeners in attach not ready

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/313


justinfagnani:
LGTM

Issue #231, opened 2015-11-02
Mobile A11y Testing 11/1


lpalmaro:
On Android with Talkback, when swiping linearly with Talkback, I first reach "text input", swipe again and hear "tap to select" and the focus is drawn around a blank space. Then I swipe again and reach the label field, which is fine. This blank focusable item also exists right after it says Text Area. This is not a huge deal, but it just seems a bit odd to swipe to something and have a blank area focused with the focus ring around it.

notwaldorf:
Oof, what seems to be happening here is that Talkback goes through *every* element on the page, regardless of whether those elements are focusable. Here's a JSBin (http://output.jsbin.com/mevore) that shows how Talkback goes through 2 `<br>` elements after the first `<input>`, the `<p>` paragraph, and then the `<textarea>`. Conversely, Voice Over only goes through  the `<input>` and the `<textarea>`, which makes sense.

This is exactly what's happening to `paper-input`: it internally has a bunch of `<divs>` (like the label), and talkback goes through all of them :(

@lpalmaro I'm not super sure how we can fix this, since this works fine on all the other voice readers, and this seems to be a behavior of just Talkback (pls correct me if I'm wrong)

notwaldorf:
Good news! I have a fix for this! 
Bad news! It's blocked on a Chrome Talkback bug: https://bugs.chromium.org/p/chromium/issues/detail?id=594791

PR #360, opened 2016-03-14
aria-hide the label since it's not needed

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/231

The label can be `aria-hidden` since it's already being read out as part of the `input`'s `aria-labelledby`.

Don't merge yet: currently blocked on https://bugs.chromium.org/p/chromium/issues/detail?id=595494


valdrinkoshi:
Tests are green! \o/
LGTM

notwaldorf:
https://bugs.chromium.org/p/chromium/issues/detail?id=595494 seems fixed (at least in M51), so this is finally ready for a merge!

PR #361, opened 2016-03-15
jscompiler doesn't like behaviors being defined in a closure

garlicnation:
I moved the global ID variables to Polymer, but I'm willing to accept any other strategy that keeps the definition of PaperInputBehaviorImpl outside a closure.


notwaldorf:
@garlicnation This needs a rebase now (I wanted to check the tests )

PR #363, opened 2016-03-18
Add missing backticks

arthurevans:
These missing backticks result in some missing text in the API docs.

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -25,7 +25,7 @@ For example:
       <input is="iron-input">
     </paper-input-container>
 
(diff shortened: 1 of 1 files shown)


notwaldorf:
@arthur would you mind fixing that 'contanter' typo while you're at it too? :)

arthurevans:
Oh, does it apply to _containers_ as well as _contanters_? 

D'oh. My motto: work fast, work dumb.

notwaldorf:
The tests are failing on Safari 7 for unrelated reasons; going to merge this in the meantime, since it's just a docs fix anyway.

PR #368, opened 2016-03-23
fixed browser back/forward input field value

jannakha:
when user moved back/forward value in the field is not saved, although standard browser behaviour is to save state/value of input fields 
fixed for Safari/Firefox (Chrome didn't have this issue)


notwaldorf:
This actually breaks `type=number`. If you try this patch on an `<paper-input type=number>` and type an invalid input (like `111eee`) the label unfloats incorrectly: http://jsbin.com/jipise/edit?html,output

I will revert this patch tomorrow :(

PR #378, opened 2016-04-11
Fix always floating label on required, empty type=number input

### LINKED ISSUE #339: paper-input type="number" the label is always floating
In case paper-input contains "required" property and type="number", the label is always floating even if value is not defined:

http://jsbin.com/kasivaxuni/edit?html

samuelli:
Fixes #339

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -515,8 +515,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
     _handleValue: function(inputElement) {
       var value = this._inputElementValue;
 
-      // type="number" hack needed because this.value is empty until it's valid
-      if (value || value === 0 || (inputElement.type === 'number' && !inputElement.checkValidity())) {
+      if (value || value === 0) {
         this._inputHasContent = true;
       } else {
         this._inputHasContent = false;


paper-input-container.html
```diff
@@ -515,8 +515,7 @@
     _handleValue: function(inputElement) {
       var value = this._inputElementValue;
 
-      // type="number" hack needed because this.value is empty until it's valid
```
notwaldorf: Hmmm, this worries me. If the input has type="number" and a junk input (like `111eee`), then this internally sets the value to the empty string: http://jsbin.com/suruje/edit?html,output

So basically, if you remove this and type the (allowed) string `111eee` into the input, it will say "oh I have no content", and display the label on top of the input.

samuelli: I might be crazy but this seems to be already broken.

Repro locally on `paper-input` `type="number"` using input `111ee` results in an empty input.

Can't repro on jsbin/don't know how to make polygit pull in the latest version.

notwaldorf: I updated the JSBin: http://jsbin.com/suruje/edit?html,output. If I type `111eee` I get: 
screen shot 2016-04-13 at 11 45 34 pm

With your change patched in, if I type `111` everything is fine, but as soon as I start typing the `e`s I get:
screen shot 2016-04-13 at 11 49 25 pm

This is because the moment you type the first `e`, internally the native `input` goes "nope, invalid", and sets a) the validity to false and b) the `value` property to `""`. This means that the code above won't run (with your change), and it will set `inputHasContent` to false, which will unfloat the label.

notwaldorf: And here's a JSBin that uses _your_ branch, for comparison: http://jsbin.com/suruje/edit?html,output

Type `111` and then an `e` in the `paper-input` to repro.

samuelli: But without my changes on master http://jsbin.com/kehimanape/1/edit?html,output its already broken I think.

I think the other jsbin pulls in a different version of something.

notwaldorf: Ooof, you're right. I wonder if it's this change: https://github.com/PolymerElements/paper-input/commit/c0fd9403ec77b4f0db67ed3526482d4dfc9f06c3 :/

samuelli: Ok, thanks!

samuelli: Can you revert that PR please?

Issue #292, opened 2015-12-23
Error message based on error type


srikkbhat:
Would be great to assign error message based on error type. I have a below scenario 

1. Input field that accepts email. 
2. Field is required. So if user submits the form without entering the value then show message that field is required.
3. Pattern should match email. Auto validate to check if pattern matches. If pattern do not match show error message that email is invalid.
4. When user submits the form, check the email in the backend. If the email is not found, show error message that email is not found.

Whenever user starts typing, I have to clear messages `#2` or `#4` and assign `#3` to errorMessage. When user submit the form then I have to assign `#2` or `#4` based on the error. 

I think there are definite advantages to have this built into the paper elements.

MeTaNoV:
did you checked [`gold-email-input`](https://elements.polymer-project.org/elements/gold-email-input) ?
And if you want additionally the password equivalent: https://github.com/MeTaNoV/gold-password-input

srikkbhat:
My request is valid for `gold-email-input` too. I thought `paper-input` is a good place to create this request.

notwaldorf:
I think that this is way too specific logic to be built in a generic element. What if you also want to change the message based on what kind of error you have? (input too short/ input super invalid/ input is a particularly evil string).

This sounds like application logic, which means it would be very hard to bake it correctly in a generic element, and sounds like it would duplicate the `ValidityState` API a lot (https://developer.mozilla.org/en-US/docs/Web/API/ValidityState).

I would do something like this (ditching the `auto-validate`, and leveraging the `ValidityState` API + your custom logic): http://jsbin.com/zizeri/edit?html,output

srikkbhat:
Thanks for the references and example. 

My request was inspired by the custom validation in angular framework https://docs.angularjs.org/guide/forms#custom-validation. We can define our own validators which will set the invalid state of the input. We can also reference it to set error messages.  The validity api seems to do most of the work.

PR #387, opened 2016-05-05
support added to set error message for each validity state

tinkerD:


notwaldorf:
I'm not super sure this should be baked into the element itself, it makes the element more complicated to use. Instead, I think it should be done by the element user. I've been asked this [before](https://github.com/PolymerElements/paper-input/issues/292#issuecomment-169169711); here's a demo of how I think this should work: http://jsbin.com/zizeri/edit?html,output

PR #392, opened 2016-06-03
Clear the focused attribute on blur from shift-tab. Fixes #386

### LINKED ISSUE #386: Focused property is not cleared when using shift+tab
### Description
The `focused` property (and corresponding attribute) is not cleared if focus is lost via navigating the page with `shift+tab` keys. When going past a `paper-input` with `tab` key, the `focused` property is correctly cleared.

### Expected outcome

The `paper-input` should not have `focused` as `true` when focus was already lost.

### Actual outcome

The `paper-input` has `focused` as `true` and the `focused` attribute present.

### Live Demo
See the animated gif with a background style bound to the `focused` attribute.
paper-input-focused

### Steps to reproduce
1. Put couple of `paper-input` elements in the page.
2. Open the page in a web browser.
3. Navigate the page with `tab` and `shift+tab` and see how `focused` attribute behaves.

### Browsers Affected
At least the following:
- [x] Chrome
- [x] Firefox
- [x] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

tehapo:
This PR fixes clearing of the `focused` attribute when the blur occurs as a result of `shift-tab` key combination. Basically this was fixed by always calling the `_focusBlurHandler` of `IronControlState` even when the `_shiftTabPressed` flag is set.

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -446,13 +446,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * Forward focus to inputElement. Overriden from IronControlState.
      */
     _focusBlurHandler: function(event) {
-      if (this._shiftTabPressed)
-        return;
-
       Polymer.IronControlState._focusBlurHandler.call(this, event);
 
       // Forward the focus to the nested input.
-      if (this.focused)
+      if (this.focused && !this._shiftTabPressed)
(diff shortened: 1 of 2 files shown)


notwaldorf:
LGTM! Thanks so much!

PR #398, opened 2016-06-22
Ensure hidden attribute works.

### LINKED ISSUE #397: hidden attribute does not work for paper-input inside shadow dom
### Description
`paper-input` does not respect the hidden attribute when used inside a shadow root.  This is a common issue in many Polymer elements, I'm just randomly starting with `paper-input`.

### Expected outcome
The `hidden` attribute should cause `display:none` to apply.

### Actual outcome
The `:host` selector takes over, causing `paper-input` to remain visible even with hidden attribute.

### Live Demo
https://jsbin.com/gukazaxenu/edit?html,output

### Proposed fix
We should be using `:host(:not([hidden]))` selector to specify the `display` option of custom elements.

coreyfarrell:
Fixes #397.

### DIFF
--- a/demo/ssn-input.html
+++ b/demo/ssn-input.html
@@ -20,6 +20,10 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         display: inline-block;
       }
 
+      :host([hidden]) {
+        display: none !important;
+      }
+
       input[is="iron-input"] {
         font: inherit;
         outline: none;
--- a/paper-input-char-counter.html
+++ b/paper-input-char-counter.html
@@ -41,6 +41,10 @@ Custom property | Description | Default
         @apply(--paper-input-char-counter);
       }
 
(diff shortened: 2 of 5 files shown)


demo/ssn-input.html
```diff
@@ -20,6 +20,10 @@
         display: inline-block;
       }
 
+      :host([hidden]) {
+        display: none!important;
```
notwaldorf: Sorry to be a nitpicker -- would you mind adding a space before the `!important`, please?

notwaldorf:
Looks good -- just a small nit left

demo/ssn-input.html
coreyfarrell: No problem, I understand the value of uniform code style.  Updated commit running through CI now.

notwaldorf:
Thanks so much! 👍

PR #400, opened 2016-06-24
Fixes #394

### LINKED ISSUE #394: Paper-textarea label is offset from underline
### Description
Paper-textarea label is offset 4 pixels above the underline whereas paper-input is not offset at all. If they are meant to appear differently, there probably needs to be a more substantial difference.

### Expected outcome
Both paper-input and paper-textarea appear similar before any text is entered

### Actual outcome
Paper-input and paper-textarea appear differently.

### Live Demo
https://jsbin.com/dayedifeqo/edit?html,output

### Steps to reproduce
Add paper-input, give label.
Add paper-textarea, give label.

kreinholt:
Adds vertical-align:bottom to input. Only an issue when doctype is declared. Fixes #394

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -272,6 +272,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
         color: var(--paper-input-container-input-color, --primary-text-color);
         -webkit-appearance: none;
         text-align: inherit;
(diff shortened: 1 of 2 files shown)


kreinholt:
superseded by #439

PR #401, opened 2016-06-24
If autofocus is set to true, implicitly take focus. Fixes #166.

### LINKED ISSUE #166: Autofocus not working in Firefox
Firefox seems to evaluate `autofocus` only at page load. As this attribute is applied by `paper-input` after page load, firefox never takes notice of it.

Here's a [JSBin](http://output.jsbin.com/camosonafo)
It show's the raw output, because autofocus seems not to work at all in the edit mode of JSBin.

Cheers

JanMiksovsky:
The fix here was made unexpectedly challenging by some odd activeElement results in IE 11.

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -212,7 +212,8 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
        * element, bind this to the `<input is="iron-input">`'s `autofocus` property.
        */
       autofocus: {
-        type: Boolean
+        type: Boolean,
+        observer: '_autofocusChanged'
       },
 
       /**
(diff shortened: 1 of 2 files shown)


paper-input-behavior.html
```diff
@@ -529,6 +530,33 @@
...
+        // outermost html element, but there are also cases (under the
+        // polyfill?) in which the activeElement is not a real HTMLElement, but
+        // just a plain object. We identify the latter case as having no valid
+        // activeElement.
+        var activeElement = document.activeElement;
+        var isActiveElementValid = activeElement instanceof HTMLElement;
+
+        // Has some other element has already taken the focus?
+        var isSomeElementActive = isActiveElementValid &&
+            activeElement !== document.body &&
```
notwaldorf: @cdata and I were talking, and I think this may be a problem if someone adds a shadow root to `body`. It's very unlikely though, so it should be fine.

notwaldorf:
LGTM!

PR #406, opened 2016-07-02
Remove default-theme.html that is for demo

JosefJezek:


notwaldorf:
The default theme is not for the demo -- it's used to define the default colours of the element (like the grey underline and text, and indigo colour when focused)

Issue #408, opened 2016-07-11
A11y: Upon showing an error using error-message attribute no verbalisation is given


giltza:
### Description
Error messages that are shown in response to invalid = true && error-message="<text here>" are not verbalized when using accessibility tools.

### Live Demo
https://jsfiddle.net/a3msv0tu/

### Steps to reproduce

1. Turn on a verbalisation acceability tool (ChromeVox/Voice Over/etc.)
2. Update using code the paper-input to invalid="true" and error-message="This is an error message"

### Expected
Accessibility tool should 'speak' the error message that appears.

### Actual result
The error message is not 'Spoken' 

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

notwaldorf:
Fixed in https://github.com/PolymerElements/paper-input/pull/409

PR #409, opened 2016-07-11
Fixing #408: A11y: Upon showing an error using error-message attribute no verbalisation is given.

giltza:
Fixing #408
Adding aria-live="assertive" to the error message in paper-input.


giltza:
Hi @notwaldorf,

Can you please take a look?

Thanks

notwaldorf:
This works great, thanks so much!

PR #419, opened 2016-08-02
no pointer events for label

### LINKED ISSUE #418: tap event fired twice with dom=shadow
### Description

When in `dom="shadow"`, tapping on the label will trigger 2 tap events. The label should be inert/have no pointer events.

### Expected outcome

Only one tap event should be fired

### Actual outcome

Two tap events are fired

### Live Demo
http://jsbin.com/zufito/4/edit?html,console,output

Originally reported in https://bugs.chromium.org/p/chromium/issues/detail?id=633496

valdrinkoshi:
Fixes #418 by disabling pointer events for `label` in `paper-input` and `paper-textarea`.
This is not done in `paper-input-container` as it might remotely be a breaking change for whoever is counting on pointer events on the label. Moreover, users of `paper-input-container` can already disable pointer events by using the mixin `--paper-input-container-label`

### DIFF
--- a/paper-input.html
+++ b/paper-input.html
@@ -73,7 +73,7 @@ style this element.
       :host {
         display: block;
       }
-      
+
       :host([focused]) {
         outline: none;
       }
@@ -96,6 +96,10 @@ style this element.
 
       input:-ms-input-placeholder {
         color: var(--paper-input-container-color, --secondary-text-color);
+      }
+
+      label {
+        pointer-events: none;
       }
     </style>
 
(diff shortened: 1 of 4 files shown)


notwaldorf:
Maybe do this for `paper-textarea` too, to be symmetric?

LGTM otherwise.

valdrinkoshi:
Good catch! Updated `paper-textarea` as well

notwaldorf:
👍

PR #430, opened 2016-08-24
Add missing binding for invalid property to paper-textarea.

krozycki:
This prevents the invalid property from getting out of sync when it it for example set from an outside component.


fejesjoco:
The same binding is there in paper-input for iron-input. Not having this binding causes bugs when the invalid value doesn't propagate. So this should be a safe and important change.

notwaldorf:
LGTM, thanks so much!

PR #435, opened 2016-09-06
Added mixins to style with different invalid colors

gerjanvangeest:
I wanted to use the paper-input but not style the label, underline and add-on in the same color when invalid. In the current situation all three elements where all styled by the variable:
--paper-input-container-invalid-color

With these three new mixins the label, underline and add-on can be styled individually.


gerjanvangeest:
I signed the CLA!

notwaldorf:
This is pretty against the material design spec, and it would produce a pretty mismatched input, so I'm not sure how I feel about this :(

Issue #289, opened 2015-12-21
How to apply multiple validator to a single paper-input in polymer?


maxiplay:
I don't know if its is a fonctionnality miss or a choice.

My question.

http://stackoverflow.com/questions/34401651/how-to-apply-multiple-validator-to-a-single-paper-input-in-polymer

notwaldorf:
That is not currently supported at the moment.

jimmyvithalani:
Hi,
I am interested in this feature and would like to know when the feature will be merged and in which version will it be available. I am currently on v1.6 of polymer and v1.0.10 of paper input.
Sorry if this is not the right place to put such a question/comment, I am new to this world so don't know the processes.
Thanks.

PR #436, opened 2016-09-08
support for multiple error messages (which are defined in validators)

tlouisse:
Currently, it's not possible to use multiple validators on an input.

Also see:
- https://github.com/PolymerElements/paper-input/issues/289
- https://github.com/PolymerElements/iron-input/issues/53

With this pull request, it will become possible to add multiple validators to elements implementing the validatable behavior, without having to write a new custom validator that combines functionality of other validators. 

Along with this functionality, the following will be supported:
- custom messages: 
	- stored in validators (so there's a single source of truth, resulting in consistency between forms and less development work)
	- support for multiple error messages, ordered by priority (the most important message will be shown by the paper-input)
- support for native validators('required', 'max', 'min', 'step', 'maxlength', 'minlength', 'pattern', 'emailtype'(type=email) and 'urltype' (type=url))
- automatic instantiating of custom validators 
- possibility to connect unique instances of a validator to a validatable-element (when overriding of the default message will be needed)

Everything is tested for backwards compatibility. All tests have run in chrome 51, firefox 46, OS X 10.9 safari 7, OS X 10.10 safari 8, OS X 10.11 safari 9, Windows 7 IE 10, Windows 8.1 IE 11, Windows 10 microsoftedge, Android 4.3, Android 4.4, Android 5, Android 6, iOS 7, iOS 8, iOS 9.3.

Note that, since the validation functionality is located in different repositories, the above functionality will contain pull requests in the following repositories:
- iron-validator-behavior
- iron-validable-behavior
- iron-input
- paper-input

The link below shows an example of what would be the benefit of the old vs. the new situation.
It makes clear what the API will look like and how little code will be needed to achieve the same in the new situation.

https://tlouisse.github.io/iron-validatable-behavior/components/iron-validatable-behavior/demo/

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -32,7 +32,7 @@
     "iron-autogrow-textarea": "PolymerElements/iron-autogrow-textarea#^1.0.0",
     "iron-behaviors": "PolymerElements/iron-behaviors#^1.0.0",
     "iron-form-element-behavior": "PolymerElements/iron-form-element-behavior#^1.0.0",
-    "iron-input": "PolymerElements/iron-input#^1.0.0",
+    "iron-input": "https://github.com/tlouisse/iron-input.git#feature/multipleValidators",
     "paper-styles": "PolymerElements/paper-styles#^1.1.4",
     "iron-a11y-keys-behavior": "PolymerElements/iron-a11y-keys-behavior#^1.0.0"
   },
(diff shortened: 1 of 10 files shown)


vospascal:
something like this would be nice now there is no way to chain validator's

<paper-input-container>
  <paper-input validator="validatorA validatorB"/>
  <paper-input-error validator-name="validatorA">validatorA message</paper-input-error>
  <paper-input-error validator-name="validatorB">validatorB message</paper-input-error>
</paper-input-container>

PR #444, opened 2016-09-26
Add CSS mixing for webkit spinner

### LINKED ISSUE #279: Disable spinner on `type='number'`
So, you can do this using CSS using

```
input[type="number"]::-webkit-outer-spin-button,
input[type="number"]::-webkit-inner-spin-button {
  -webkit-appearance: none;
  margin: 0;
}
input[type="number"] {
  -moz-appearance: textfield;
}
```

so this can only be achieved right now on Firefox with the current set up. There are two ways this could be exposed:
1. Add a Chrome (`::-webkit-inner-spin-button`) /Safari (`::-webkit-outer-spin-button`) specific mixin, thus the user will have to write something like
   
   ```
   paper-input {
     --paper-input-container-input-webkit-spinner: {
       -webkit-appearance: none;
       margin: 0;
     }
     -moz-appearance: textfield;
   }
   ```
2. Add a `no-number-spinner` attribute which applies the CSS for the user instead.

TimvdLippe:
Was facing the same issue and saw there was already an issue open for it. I have tested it in my application and it is working perfectly with:

``` css
paper-input[type="number"] {
  --paper-input-container-input-webkit-spinner: {
    -webkit-appearance: none;
    margin: 0;
  }
}
```

Fixes #279

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -90,6 +90,7 @@ Custom property | Description | Default
 `--paper-input-container-label-focus` | Mixin applied to the label when the input is focused | `{}`
 `--paper-input-container-label-floating` | Mixin applied to the label when floating | `{}`
 `--paper-input-container-input` | Mixin applied to the input | `{}`
+`--paper-input-container-input-webkit-spinner` | Mixin applied to the webkit spinner | `{}`
 `--paper-input-container-underline` | Mixin applied to the underline | `{}`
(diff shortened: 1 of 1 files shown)


TimvdLippe:
Friendly ping @notwaldorf

notwaldorf:
Thanks!

Issue #405, opened 2016-06-30
underline mixins may need tweaking (have to override 2 in order to hide underline)


valdrinkoshi:
_From @ktiedt on April 27, 2016 18:21_

### Description

In the demo it explains how to remove the underline from a paper-dropdown-menu however it doesnt work because `--paper-inout-container-underline-focus` is also not overridden, as seen here: https://github.com/PolymerElements/paper-dropdown-menu/blob/master/demo/index.html#L218-L221
### Expected outcome
Would not expect to have to override both `--paper-input-container-underline` and `--paper-input-container-underline-focus` to hide the underline

### Actual outcome
underline appears when you click the paper-dropdown-menu that is said to have no underline.

### Live Demo

https://elements.polymer-project.org/elements/paper-dropdown-menu?active=paper-dropdown-menu&view=demo:demo/index.html

_Copied from original issue: PolymerElements/paper-dropdown-menu#155_

valdrinkoshi:
Will move this issue to `paper-input`, as those are mixins of `paper-input`

notwaldorf:
It would be nice to add a demo of how to style the underline

ktiedt:
Ideally that does not involve overriding two separate css vars ;)

yaqihu:
Would it be possible to keep the underline appear when the dropdown is opened?
Right now, for paper-dropdown-menu, it disappears when the dropdown is opened, which cause a flash visually.

PR #451, opened 2016-10-22
Common style property underline, underline focused and underline disa…

aperrad:
Add common style for underline, underline when input is focused, and when it is disabled with css variable _--paper-input-underline-common_
#405


notwaldorf:
You need to sign the CLA before we can merge this :(

aperrad:
I sign it with my correct email adress. What do I need to do ?

PR #454, opened 2016-10-30
Update paper-input.html

mkazlauskas:
Fixes the following error:

error

Note: I encountered a similar error in [iron-a11y-keys-behavior](https://github.com/PolymerElements/iron-a11y-keys-behavior/pull/59). I assume `registered` callback doesn't happen at the same time as in 1.0.

### DIFF
--- a/paper-input.html
+++ b/paper-input.html
@@ -197,7 +197,7 @@ style this element.
       Polymer.PaperInputBehavior
     ],
 
-    registered: function() {
+    created: function() {
       // These types have some default placeholder text; overlapping
       // the label on top of it looks terrible. Auto-float the label in this case.
       this._typesThatHaveText = ["date", "datetime", "datetime-local", "month",


notwaldorf:
Are you trying to use this with 2.0? `registered` is very different from `created`: `registered` means it's done once per tag name (i.e. when the class is registered), whereas `created` is done once per element instance.

mkazlauskas:
Yes, with 2.0-preview. I know it's different, I assume in "legacy element" registered doesn't work the same way as in 1.0 (2.0 is probably supposed to use "static" members for this purpose). I assumed this PR is ok since it's not expensive to do in `created` and it makes it work.

notwaldorf:
`registered` should work in 2.0 but ohhhhh this uses `this`, so I think I just wrote a dumb (it actually needs to modify the element instance)

Issue #448, opened 2016-10-13
Underline disappears sporadically as we zoom out.


StokeMasterJack:
Below are three screen shots at various zoom levels in Chrome:

**Large:**
![large](https://cloud.githubusercontent.com/assets/123349/19359084/62c6fa62-912d-11e6-9ed3-8090c366dc1f.png)

**Medium:**
![medium](https://cloud.githubusercontent.com/assets/123349/19359082/62c1adf0-912d-11e6-9af3-fb1da9ea6ba4.png)

**Small:**
![small](https://cloud.githubusercontent.com/assets/123349/19359083/62c3568c-912d-11e6-93ae-91a44f0058fc.png)

freshp86:
This is also affecting Chrome's MD Settings page, see [crbug.com/663304](https://bugs.chromium.org/p/chromium/issues/detail?id=663304).

notwaldorf:
Is this on Retina or non-retina? Sounds like a subpixel rounding error :(

freshp86:
I can reproduce both on a regular Dell 24 inch display, as well as on my Chromebook Pixel.

notwaldorf:
This definitely looks like subpixel problems. In this jsbin (http://output.jsbin.com/soqevu), the bottom two lines are just divs with heights of 1px and 2px (which is what a `paper-input` uses internally for the underline). On my retina Mac
- at 90% and 33% the 1px standalone underline disappears
- at 67% the 1px paper-input underline disappears
- at 25% both disappear

notwaldorf:
Moreover, this doesn't happen at all on Safari, and on FF only the 1px standalone underline disappears at the smallest possible zoom level, so this sounds to me like a Chrome bug. Closing :(

freshp86:
If you use bottom-border instead of a div, to represent the underline, I can no longer repro. Would that be a valid fix instead of just closing this issue?

danbeam:
as explained by @eaenet in https://trac.webkit.org/wiki/LayoutUnit, borders ensure they stay the right thickness as zoom changes (`LayoutUnit`s are the underlying units used for subpixel rendering, or at least they were at a time).

@notwaldorf I'm hitting very similar issues in Firefox 49.0.2 on https://jsfiddle.net/ea2hjbz7/1/: dropping or fattening of lines implemented with `<div style="height: ...">` (but not with `border`).

I can see a few options:

[...]

PR #455, opened 2016-11-11
Implement underlines with border rather than height/width for better rendering at non-standard zoom

danbeam:
addresses https://github.com/PolymerElements/paper-input/issues/448


danbeam:
@notwaldorf @freshp86

notwaldorf:
LGTM! Thanks!

Issue #463, opened 2016-12-02
Basic and [focused] mixing for paper-input


Westbrook:
### Description
I would like to be able to use `custom-styles` to set styling across all of the `paper-input` elements in my app. There are a number of really useful ones internal to `paper-input-container` but nothing at the top level.

### Expected outcome
Basic and [focused] styles can be applied at the app level to `paper-input` elements across an application.

### Actual outcome
Included elements are needed locally to each custom element that employs the input to access `paper-input` and `paper-input[focused]`.

### Steps to reproduce
1. Attempt to share custom styles to `paper-input` across an application.
2. Fail.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari 9
- [x] Safari 8
- [x] Safari 7
- [x] Edge
- [x] IE 11
- [x] IE 10

robrez:
I was just coming here to request that focus styling be added to input-container.. we have `--paper-input-container-input`, but changing properties when the input is focused would be nice... eg: `--paper-input-container-input-focus`

I have a polymer webapp that is deployed for multiple customers, we allow full-theming per customer...  some have requested more "traditional" looking inputs...

![focus](https://cloud.githubusercontent.com/assets/1993389/20851137/26f486e6-b8a4-11e6-9f59-db16a1917a2a.png)

notwaldorf:
Regarding the original request: I think that's more complicated than it sounds, because each individual element inside the `paper-input` is styled separately when the element is focused, so the mixin wouldn't really affect that.

There's already `--paper-input-container-focus-color` to change the colour, and mixins for `--paper-input-container-underline-focus` and `--paper-input-container-label-focus`. Is the only thing you're missing then, is `--paper-input-container-input-focus`?

Westbrook:
It's possible that `--paper-input-container-input-focus` would achieve a similar goal as to what I'm looking to do. More specifically, I have the following file: 
```
<dom-module id="comps-input-styles">
  <template>
    <style>
	paper-input {
		display: block;
		box-sizing: border-box;
		border: solid 1px gray;

		--paper-input-container: {
			border: none;
			padding: 15px 20px;
		};

		--paper-font-subhead: {
			font-size: 100%;
		};

		--paper-input-container-color: rgb(64, 64, 64);
		--paper-input-container-focus-color: rgb(64, 64, 64);

		--paper-input-container-underline-focus: {
			display: none;
		};

		--paper-input-container-underline: {
			display: none;
		};

[...]

PR #464, opened 2016-12-02
Implementat @apply mixins to control styling

Westbrook:
Surface default (`:host`) and focused (`:host([focused])`) targeted mixin implementations for the component.

references #463

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -90,6 +90,8 @@ Custom property | Description | Default
 `--paper-input-container-label-focus` | Mixin applied to the label when the input is focused | `{}`
 `--paper-input-container-label-floating` | Mixin applied to the label when floating | `{}`
 `--paper-input-container-input` | Mixin applied to the input | `{}`
+`--paper-input-container-input-focus` | Mixin applied to the input | `{}`
+`--paper-input-container-input-invalid` | Mixin applied to the input | `{}`
 `--paper-input-container-input-webkit-spinner` | Mixin applied to the webkit spinner | `{}`
 `--paper-input-container-underline` | Mixin applied to the underline | `{}`
 `--paper-input-container-underline-focus` | Mixin applied to the underline when the input is focused | `{}`
@@ -275,6 +277,20 @@ This element is `display:block` by default, but you can set the `inline` attribu
 
         @apply(--paper-font-subhead);
         @apply(--paper-input-container-input);
+      }
+
+	  .input-content ::content input:focus,
+      .input-content ::content textarea:focus,
+      .input-content ::content iron-autogrow-textarea:focus,
+      .input-content ::content .paper-input-input:focus {
+        @apply(--paper-input-container-input-focus);
+      }
+
+	  .input-content ::content input[invalid],
+      .input-content ::content textarea[invalid],
+      .input-content ::content iron-autogrow-textarea[invalid],
+      .input-content ::content .paper-input-input[invalid] {
+        @apply(--paper-input-container-input-invalid);
       }
 
       .input-content ::content input::-webkit-outer-spin-button,
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -279,17 +279,17 @@ This element is `display:block` by default, but you can set the `inline` attribu
         @apply(--paper-input-container-input);
       }
 
-      .input-content ::content input:focus,
-      .input-content ::content textarea:focus,
-      .input-content ::content iron-autogrow-textarea:focus,
-      .input-content ::content .paper-input-input:focus {
+      .input-content.focused ::content input,
+      .input-content.focused ::content textarea,
+      .input-content.focused ::content iron-autogrow-textarea,
+      .input-content.focused ::content .paper-input-input {
         @apply(--paper-input-container-input-focus);
       }
 
-      .input-content ::content input[invalid],
-      .input-content ::content textarea[invalid],
-      .input-content ::content iron-autogrow-textarea[invalid],
-      .input-content ::content .paper-input-input[invalid] {
+      .input-content.is-invalid ::content input,
+      .input-content.is-invalid ::content textarea,
+      .input-content.is-invalid ::content iron-autogrow-textarea,
+      .input-content.is-invalid ::content .paper-input-input {
         @apply(--paper-input-container-input-invalid);
       }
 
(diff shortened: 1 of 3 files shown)


notwaldorf:
As I mentioned in #463, I don't think this is the correct solution, since it won't really do much (it will control the outside of the element, but nothing like the label, the input or the underline)

paper-input-container.html
```diff
@@ -90,6 +90,8 @@
 `--paper-input-container-label-focus` | Mixin applied to the label when the input is focused | `{}`
 `--paper-input-container-label-floating` | Mixin applied to the label when floating | `{}`
 `--paper-input-container-input` | Mixin applied to the input | `{}`
+`--paper-input-container-input-focus` | Mixin applied to the input | `{}`
```
notwaldorf: Please change this comment to 'Mixin applied to the input when focused', like the `-label-focus` mixin is commented. Similar comment for the `-input-invalid` one

paper-input-container.html
```diff
@@ -277,6 +279,20 @@
         @apply(--paper-input-container-input);
       }
 
+	  .input-content ::content input:focus,
```
notwaldorf: weird indentation here -- left align all the things :)

paper-input-container.html
```diff
@@ -277,6 +279,20 @@
...
+      .input-content ::content textarea:focus,
+      .input-content ::content iron-autogrow-textarea:focus,
+      .input-content ::content .paper-input-input:focus {
+        @apply(--paper-input-container-input-focus);
+      }
+
+	  .input-content ::content input[invalid],
```
notwaldorf: I think this should be `.input-content.is-invalid ::content input` and friends (these friends, in particular: https://github.com/PolymerElements/paper-input/blob/master/paper-input-container.html#L259) , like we use it here: https://github.com/PolymerElements/paper-input/blob/master/paper-input-container.html#L249

paper-input-container.html
notwaldorf: Also this should use something about `this.focused`, so maybe `:host([focused]) .input-content ::content input`? (we use `focused` to style the floating label in the computed class)

paper-input-container.html
Westbrook: This seems like it would be a good idea, but if the input becomes "invalid" due to being required while being empty, `_computeInputContentClass` will skip the addition of `is-invalid` because `_inputHasContent` would be false. https://github.com/PolymerElements/paper-input/blob/master/paper-input-container.html#L581 What are you're thoughts on this?

paper-input-container.html
Westbrook: Currently in `this.focused` notifies, but doesn't `reflectToAttribute` until `paper-input`. We don't currently have access to `:host-context([focused])`, right? Would this be something that could be added to `_computeInputContentClass`?

Westbrook:
@notwaldorf do you have any thoughts on the notes I've left around "invalid" and "focused" selectors? I'd love to wrap up the PR for you!

notwaldorf:
I am not convinced the `[invalid]` attribute works -- for example, if your `paper-input` uses a validator, _it_ can be `invalid` despite the nested `input` element being perfectly valid :(

paper-input-container.html
notwaldorf: We can maybe use a computed class, then, like we do for the underline and friends? `[[_computeUnderlineClass(focused,invalid)]]`

Westbrook:
Absolutely, that's why I'm in favor of both of your notes, I just wanted to make sure that my understanding of the side effects were correct before I started in on actual JS changes so that those notes could be fully supported. If you are good with my comments from before I'll get those changes into the review for you to checkout ASAP.

notwaldorf:
Oh yeah, i think we can add it to that class! Sorry, the GitHub UI hid your comment and I missed it :)

(I would stay away from `host-context` since it's performance is quite bad)

Westbrook:
Committed to make sure tests were passing. I'll look into the right tests to add for the two class additions if this looking good on your end.

Westbrook:
@notwaldorf happy new year! happy new tests to watch on the changes herein. also some white space changes. let me know what you think.

paper-input-container.html
```diff
@@ -594,12 +610,21 @@
...
       } else {
         if (_inputHasContent) {
           cls += ' label-is-hidden';
         }
+        if (invalid) {
+          cls += ' is-invalid';
+        }
       }
+	  if (focused) {
+        cls += ' focused';
```
notwaldorf: Is the indentation here messed up?

paper-input-container.html
```diff
@@ -594,12 +610,21 @@
...
           }
+          if (invalid) {
+            cls += ' is-invalid';
+          }
         }
       } else {
         if (_inputHasContent) {
           cls += ' label-is-hidden';
         }
+        if (invalid) {
```
notwaldorf: Should this go in the `inputHasContent` `if` statement directly above?

paper-input-container.html
```diff
@@ -277,6 +279,20 @@
         @apply(--paper-input-container-input);
       }
 
+      .input-content.focused ::content input,
+      .input-content.focused ::content textarea,
+      .input-content.focused ::content iron-autogrow-textarea,
+      .input-content.focused ::content .paper-input-input {
+        @apply(--paper-input-container-input-focus);
```
notwaldorf: 👍

notwaldorf:
Looking good! Small questions left

paper-input-container.html
Westbrook: I put this here to cover the case that the input is invalid because it's empty. When I required field was empty the `_inputHasContent` check was preventing the error state.

Westbrook: Maybe there's something extra to go with that?

notwaldorf: Hmm, does that mean that a `required` empty field starts off invalid? That isn't correct :/

Westbrook: In my testing, it only moves to "invalid" if it has `auto-validate` and is blurred or it has `validate()` manually called against it. Do you have a context that I should check in on/add a test for in which you think this would occur?

Westbrook: @notwaldorf just checking to see if you have any other thoughts on this. Not sure how to/if I need to alter this to make sure the validation state is managed appropriately.

notwaldorf: Patching it locally and taking a look now! Thanks for your patience! ❤️

notwaldorf:
I tried this locally, and i think it looks right. Thanks for putting up with this! <3

PR #460, opened 2016-11-23
label highlight should still be applied if the label doesn't float

TomK:
As per this file's comments:
```
`--paper-input-container-focus-color` | Label and underline color when the input is focused | `--primary-color`
```
Currently this is never applied due to missing `label-is-highlighted` class on `#labelAndInputContainer` where `[label-no-float]` is applied to the host.

This PR always applies `label-is-highlighted` class to `#labelAndInputContainer` if the input is focused, regardless of float state.

PR #465, opened 2016-12-06
label highlight should still be applied if the label doesn't float

TomK:
PR to master branch as requested by @notwaldorf (see #460)
The issue exists in both 1.7 and 2.0-preview

As per this file's comments:
```
`--paper-input-container-focus-color` | Label and underline color when the input is focused | `--primary-color`
```
Currently this is never applied due to missing `label-is-highlighted` class on `#labelAndInputContainer` where `[label-no-float]` is applied to the host.

This PR always applies `label-is-highlighted` class to `#labelAndInputContainer` if the input is focused, regardless of float state.

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -586,8 +586,6 @@ This element is `display:block` by default, but you can set the `inline` attribu
 
           if (invalid) {
             cls += ' is-invalid';
-          } else if (focused) {
-            cls += " label-is-highlighted";
           }
         } else {
           // When the label is not floating, it should overlap the input element.
@@ -599,6 +597,9 @@ This element is `display:block` by default, but you can set the `inline` attribu
         if (_inputHasContent) {
           cls += ' label-is-hidden';
         }
+      }
+      if (focused) {
+        cls += " label-is-highlighted";
       }
       return cls;
     },


notwaldorf:
@TomK I think the test failures are legit, because you're skipping the `_hasContent` check

TomK:
Ah! yes sorry, I wanted to discuss this as I found it a little misleading.

The documentation states
```
`--paper-input-container-focus-color` | Label and underline color when the input is focused | `--primary-color`
```

However that is not currently true.  The input is focused, but the class `label-is-highlighted` is only applied if the element has focus AND has content.

Our use case is to have the label (not floated) dimmed until the user focuses the input. Simply put:
```
        --paper-input-container-color: var(--primary-800-text-secondary-color);
        --paper-input-container-focus-color: var(--primary-800-text-color);
```

What is the intended effect of focus on paper-input-container?
If `--paper-input-container-focus-color` should only be applied if `_inputHasContent == true` then the documentation should be updated to reflect this and we can close this PR. (We will have to find another way to handle our use case)
Otherwise, I can update the tests to suit the documented behaviour and ensure that this PR passes them.

Let me know how you'd like to proceed :)

Thanks,
Tom

TomK:
FWIW, I've managed to achieve this in the host element with the following hack:
```
      :host paper-input {
        color: var(--fort-list-filter-color);
        --primary-color: var(--fort-list-filter-color);
        --primary-text-color: var(--fort-list-filter-color);
        --secondary-text-color: var(--fort-list-filter-color);
      }

      :host paper-input:focus,
      :host paper-input[focused] {
        color: var(--fort-list-filter-focus-color);
        --primary-color: var(--fort-list-filter-focus-color);
        --primary-text-color: var(--fort-list-filter-focus-color);
        --secondary-text-color: var(--fort-list-filter-focus-color);
      }
```

`color` is required to colour an icon we have slotted into `prefix`.
`--primary-color` is required to colour the focused line.
`--*-text-color` is required to colour the input and label text

notwaldorf:
I just re-read the issue and the PR and I'm not sure I understand anymore what the repro case is.
Here's a jsbin where I have a whole bunch of combinations of floating or non-floating labels, and 
`--paper-input-container-focus-color is always applied: http://jsbin.com/xesehov/edit?html,output

TomK:
It's been a while, I'll have another look today and get back to you.

TomK:
Hi, sorry it's taken so long to have a look at this.

Using the jsbin you posted:
If you focus the last item (floated label), the label turns green.
If you focus the the other items (not floated label), the label does not change colour.

The inline documentation says `Label and underline color when the input is focused`, but currently it is more accurately `Label and underline color when the input is focused AND the label is floated`

The PR is to make the all the labels in your jsbin turn green when focused.

TomK:
closing as this is now very stale

PR #476, opened 2017-01-23
Add missing semicolon. Fixes #475

### LINKED ISSUE #475: Missing semicolon can occasionally break uglify
### Description
This [line](https://github.com/PolymerElements/paper-input/blob/master/paper-input-behavior.html#L561) should have semicolon to prevent minify issues (can be reproduced by using `uglify` included by`vulcanize`)
 
Will send a PR to fix that. 

BTW, what about adding `eslint` to polymer elements repos?

web-padawan:


web-padawan:
@notwaldorf thanks for merging this! Would you tag these changes so we can use it in our project?

notwaldorf:
oops, of course!

PR #489, opened 2017-03-09
The tabIndex property on HTMLElement is camelCase

rictic:
(assuming that there's not also a tabindex property coming in from a behavior or something)

Found while developing the new linter.


notwaldorf:
I don't think this is correct -- tabindex is a native property, on all elements...

notwaldorf:
Uhoh -- i wonder how this ever worked?

rictic:
:ghost: it is a mystery :ghost:

PR #492, opened 2017-03-11
Fix nomenclature of CSS var. Keeping the pattern.

willianson:


notwaldorf:
Hmm, why is this PR needed? First, it's a breaking change, which we'd like to avoid if possible. Second, this doesn't match the rest of the vars (invalid-color, input-color).

Closing, as I don't think this PR is necessary. Please reopen if you have a description or some reason why this is needed

PR #495, opened 2017-03-14
Fix spinner mixin and add webkit-clear mixin

wq9:
The mixin `--paper-input-container-input-webkit-spinner` is in `paper-input.html` because the input element is there. 

Add `--paper-input-container-input-webkit-clear` mixin to be able to remove the clear button on input element.

PR #503, opened 2017-04-04
Add ::-ms-clear mixin

### LINKED ISSUE #501: Add mixin for ::-ms-clear pseudo-element
IE automatically adds a clear button to text inputs. We are using a custom clear button to enable this functionality in other browsers, and would like to use our clear button in IE as well (it has additional functionality compared to the one IE inserts).

Microsoft provides a CSS pseudo-element [::-ms-clear](https://developer.mozilla.org/en-US/docs/Web/CSS/::-ms-clear) to style this clear button, and the common way to hide it is to use the following CSS:

```
::-ms-clear {
    display: none;
    width: 0px;
    height: 0px;
}
```

I was able to hide the clear button with the following:

```
<style>
#input::shadow input::-ms-clear {
    display: none;
    width: 0;
    height: 0;
}
</style>
<div>
...
<paper-input id="input"></paper-input>
...
</div>
```

but since the ::shadow pseudo-element is deprecated I want to avoid using it. This can be fixed by adding a CSS mixin. Is there another, simpler solution that I'm missing? If not, I would be happy to open a PR adding the mixin if there is any interest from the devs.

z2oh:
This should close #501. I used #495 as a reference.

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -92,6 +92,7 @@ Custom property | Description | Default
 `--paper-input-container-input` | Mixin applied to the input | `{}`
 `--paper-input-container-input-webkit-spinner` | Mixin applied to the webkit spinner | `{}`
 `--paper-input-container-input-webkit-clear` | Mixin applied to the webkit clear button | `{}`
+`--paper-input-container-ms-clear` | Mixin applied to the Internet Explorer clear button | `{}`
 `--paper-input-container-underline` | Mixin applied to the underline | `{}`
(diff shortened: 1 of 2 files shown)


z2oh:
I just signed the CLA as specified above.

z2oh:
@notwaldorf Any change of getting this merged? I'm using a hotfix currently by styling with
```
#input::shadow input::-ms-clear {}
```
but the `::shadow` pseudo-element is deprecated so I want to move away from it if possible.

notwaldorf:
Thanks!

PR #511, opened 2017-04-12
Make _beforeRegister idempotent

TomK:
Fixes paper-input running against polymer#master.

This issue is occurring because `_beforeRegister` is currently called twice. Once the `registered` method is removed this problem goes away. We can't remove `registered` yet because the commit reintroducing `beforeRegister` has not been versioned.

This PR simply checks if the inputPlaceholder has been replaced already by checking for its existence.

### DIFF
--- a/paper-input.html
+++ b/paper-input.html
@@ -262,10 +262,12 @@ style this element.
 
     _beforeRegister: function() {
       var template = Polymer.DomModule.import('paper-input', 'template');
-      var version = Polymer.Element ? 'v1' : 'v0';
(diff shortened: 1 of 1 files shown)


TomK:
For reference:
image

TomK:
closing, @notwaldorf resolved this via 7a5a24815717cdc05643ebc9d52d103000b45642

notwaldorf:
@TomK Heh sorry about that. Fixed it as quickly as I could :)

PR #519, opened 2017-04-18
Css mixins moved to bottom of the rules block

swypse:


notwaldorf:
Uhhhh I am worried that this might actually break people :(

notwaldorf:
Is there a reason for this PR? Like is it breaking something for you, or just to clear it up (to be clear: it is a mistake they are at the bottom, but i am worried people are relying on these properties winning)

PR #526, opened 2017-04-27
2.0-preview fix

mkazlauskas:
This is a fix that allows extending this element.

Basically, registering an original `<paper-input>` processes the template and registering a superclass will fail since `template.content.querySelector('#template-placeholder')` will not find anything after it's been already processed.


notwaldorf:
This was merged as #536

PR #536, opened 2017-05-22
Fix for extending paper-input

mkazlauskas:


notwaldorf:
Thanks for the PR! I think there's something wrong with your commit history -- maybe it needs a rebase? There shouldn't be any commits that are already merged in master in it...

mkazlauskas:
I believe I managed to it (a rebase). Please confirm it's good now.

notwaldorf:
Can you explain a little what this is fixing? When would `inputPlaceholder` not exist (it's part of the template)

mkazlauskas:
Sure. My element does `extends customElements.get('paper-input')`, which inherits `beforeRegister` and the template. What `beforeRegister` does is replaces the `inputPlaceholder` with something else, so here's what happens:
1. paper-input processes the template in it's beforeRegister
2. my-input tries to process the template that's already been processed by paper-input

notwaldorf:
Makes sense, thanks!

Issue #529, opened 2017-05-03
paper-textarea always selects text on focus


feuerste:
### Description
The `paper-textarea` element always selects its text after focusing.

### Expected outcome
It does not select the text, or at least is configurable to switch off this behavior.

### Actual outcome
The text is always selected.
This is caused by `this._focusableElement.select();` inside `_focusBlurHandler` of `paper-input-behavior`. The same behavior can actually be seen, when using a `paper-dropdown-menu`, which also includes the `paper-input-behavior`.

### Live Demo
http://plnkr.co/edit/XOOY5r5jzMIglTdbphqX?p=preview

### Steps to reproduce
1. Open above live demo page.
2. Click onto "click textarea" to see the default textarea behavior.
3. Click onto "click paper-textarea" to see the behavior of the paper-textarea.

### Browsers Affected
- [x] Chrome
- [x] Firefox
- [x] Safari 9
- [x] Edge
- [x] IE 11

myfreeweb:
And `paper-input` as well! This is infuriating. Every time I want to put my cursor in some particular place it selects all the text.

feuerste:
One workaround, which however only works for the `paper-input` elements, but not directly for e.g. `paper-dropdown-menu`, is to set `elem.inputElement.textarea.select = undefined;`

notwaldorf:
Oh, sorry y'all. I did this on purpose, because I thought it was useful: https://github.com/PolymerElements/paper-input/blob/master/paper-input-behavior.html#L450

I can send a PR to remove it

myfreeweb:
Please just make it optional! It can be useful but was totally unexpected as a default.

notwaldorf:
I think optional is a bit of overkill right now (it would require a new property, and input already has a billion). I'll just remove it.

PR #537, opened 2017-05-22
don't autoselect text on click

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/529


notwaldorf:
@valdrinkoshi The focus tests fail on Travis, and have always failed (the two variants are ran in parallel and steal focus from each other)

PR #549, opened 2017-06-22
Pipe label through to aria-labels

robdodson:
Pipe label through as `aria-label`. This (i think) removes the need for the `updateAriaLabelledby` code. Specially this change is intended to fix an issue with paper-textarea. Because there is a shadow boundary between the textarea and paper-textarea, using aria-labelledby doesn't work. This PR depends on https://github.com/PolymerElements/iron-autogrow-textarea/pull/108 being merged first.

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -397,8 +397,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     },
 
     attached: function() {
-      this._updateAriaLabelledBy();
-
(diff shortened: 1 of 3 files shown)


robdodson:
@notwaldorf PTAL

notwaldorf:
@robdodson paper-textarea tests are red

notwaldorf:
@robdodson ping ping ping? let's get this boat merged!

robdodson:
@notwaldorf I'd really like to help but I don't think I have time to dig into all of these test failures :\

PR #557, opened 2017-07-28
New: Add selectionStart and selectionEnd

stramel:
- Added `selectionStart` and `selectionEnd` properties to better mimic `textarea` and allow it to really function as a replacement.
- Fixed a few lint errors

@notwaldorf Seems like you have the most commits. Figured you might be able to review.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,5 @@
 language: node_js
-sudo: required
-before_script:
-  - npm install -g polymer-cli
-  - polymer install --variants
+sudo: false
 env:
   global:
     - secure: >-
@@ -12,15 +9,15 @@ env:
 node_js: stable
 addons:
   firefox: latest
-  apt:
-    sources:
-      - google-chrome
-    packages:
-      - google-chrome-stable
+  chrome: stable
   sauce_connect: true
+before_script:
+  - polymer lint
+install:
+  - npm install -g polymer-cli
+  - polymer install --variants
 script:
-  - xvfb-run polymer test -l chrome
-  - xvfb-run polymer test -l firefox
+  - xvfb-run polymer test
   - >-
     if [ "${TRAVIS_PULL_REQUEST}" = "false" ]; then polymer test -s 'default';
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: node_js
-sudo: false
+sudo: required
+before_script:
+  - npm install -g polymer-cli
+  - polymer install --variants
 env:
   global:
     - secure: >-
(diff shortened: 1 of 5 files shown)


btopro:
+1. This allows events of what text is highlighted to bubble up correctly for start and end positions. as @stramel mentioned this would help allow it to be that much closer to a direct replacement for the `textfield` primative

stramel:
@notwaldorf @e111077 Would one of you two mind looking at this for me? :smile:

paper-textarea.html
```diff
@@ -96,18 +96,18 @@
 
     behaviors: [
       Polymer.PaperInputBehavior,
-      Polymer.IronFormElementBehavior
+      Polymer.IronFormElementBehavior,
```
notwaldorf: What's up with all these trailing commas?

.travis.yml
```diff
@@ -1,8 +1,5 @@
 language: node_js
-sudo: required
-before_script:
-  - npm install -g polymer-cli
-  - polymer install --variants
+sudo: false
```
notwaldorf: Why is this also changing as part of this PR? We prefer keeping all the `travis.yml` files consistent across all elements, so if one needs to change, they all should, via a `tedium` pass. In any case, it doesn't belong in this PR

bower.json
```diff
@@ -70,7 +70,7 @@
         "iron-validator-behavior": "PolymerElements/iron-validator-behavior#^1.0.0",
         "paper-icon-button": "PolymerElements/paper-icon-button#^1.0.0",
         "test-fixture": "PolymerElements/test-fixture#^1.0.0",
-        "web-component-tester": "^4.0.0",
+        "web-component-tester": "^5.0.0",
```
notwaldorf: do you _require_ version 5? Nothing in this PR would say you do

polymer.json
```diff
@@ -0,0 +1,5 @@
+{
```
notwaldorf: Same comment as above

paper-textarea.html
stramel: https://google.github.io/styleguide/jsguide.html#features-arrays-trailing-comma

Can revert it if you prefer

.travis.yml
stramel: I will revert and bring up for a tedium pass instead.

bower.json
stramel: Forgot to revert, I do this in testing so I can use `.only`

polymer.json
stramel: Will remove.

.travis.yml
stramel: https://github.com/PolymerLabs/tedium/issues/57

stramel:
@notwaldorf I made all changes requested with a comment about the trailing commas and a PR in Tedium to handle the additional updates. Let me know if there is anything else I need to do

stramel:
Ping @notwaldorf

stramel:
@notwaldorf Can you PTAL please?

notwaldorf:
Thanks! Sorry for the delay :/

stramel:
@notwaldorf No worries! Thank you for the feedback and getting it merged!

PR #564, opened 2017-08-30
This fixes #563 by providing a style mixin to override ::-ms-reveal for custom styles.

### LINKED ISSUE #563: Add mixin for ::-ms-reveal pseudo-element [similar to #501]
IE automatically adds a reveal button to password inputs. We are using a custom reveal button to enable this functionality in all browsers, and would like to use our reveal button in IE as well.

Microsoft provides a CSS pseudo-element ::-ms-reveal to style this clear button, and the common way to hide it is to use the following CSS:

```css
::-ms-reveal {
  display: none;
}
```

We need a simple way to hide ::-ms-reveal for custom CSS implementations.

### Browsers Affected
* Edge
* IE 11
* IE 10

ShaggyDude:
This fixes #563 by providing a style mixin to override ::-ms-reveal for custom styles.

### DIFF
--- a/paper-input.html
+++ b/paper-input.html
@@ -70,19 +70,23 @@ style this element.
 --paper-input-container-ms-clear  | Mixin applied to the Internet Explorer reveal button -aka the eyeball | {}
 
 To hide the default ms clear icon and provide your own
-`--paper-input-container-ms-clear: {
-  display: none;
-}`
+
+paper-input{
+  --paper-input-container-ms-clear: {
+    display: none;
+  };
+}
 
 To hide the default ms reveal icon and provide your own
+
 --paper-input-container-ms-reveal | Mixin applied to the Internet Explorer reveal button -aka the eyeball | {}
-`--paper-input-container-ms-reveal: {
-  display: none;
-}`
+paper-input{
+  --paper-input-container-ms-reveal: {
+    display: none;
+  }
+}
 
 -->
-
-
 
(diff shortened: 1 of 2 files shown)


ShaggyDude:
I signed it!

ShaggyDude:
CI build fails for some reason. Let me know if there is something I need to fix.

ShaggyDude:
Help!
How do I make this happen?

jean-maxime:
Can you add the documentation in paper-input-container.html like this :
`--paper-input-container-ms-reveal` | Mixin applied to the Internet Explorer clear reveal button | `{}`

notwaldorf:
yup, please document this mixin :)

ShaggyDude:
Done. Lemme know if it looks ok.

paper-input.html
```diff
@@ -64,6 +64,28 @@
 @element paper-input
 @hero hero.svg
 @demo demo/index.html
+
+### mixins
+
+--paper-input-container-ms-clear  | Mixin applied to the Internet Explorer reveal button -aka the eyeball | {}
```
notwaldorf: I think just adding this under `### Styling` without the example is enough

notwaldorf:
I'll merge this and just fix my comment in master :)

paper-input.html
ShaggyDude: Awesome thanks,
Still learning the process

PR #568, opened 2017-09-12
fix incorrect var name

craPkit:
Just found this when opening the file in an _eslint_'ed editor.


craPkit:
I signed it!

notwaldorf:
this sees to break the "error message is displayed" test?

PR #569, opened 2017-09-12
Fix a typo in the documentation of paper-input-container.

shk3:
There are missing quotes around an HTML tag in paper-input-container documentation, which is causing the HTML tag be interpreted on browsers.


shk3:
Signed.

notwaldorf:
Thanks!

PR #585, opened 2017-11-01
master - fixes #352 safari hack: focusing input w/ 0 height

### LINKED ISSUE #352: type="date" not working on Safari/iPad
There is an issue on iPad's Safari with `<paper-input type="date" />`. It is only touchable if there is an initial value set. If there is no value or the value gets deleted the field is no longer touchable. Same for `<paper-input type="time" />`.

Browser: Mozilla/5.0 (iPad; CPU OS 9_2_1 like Mac OS X) AppleWebKit/601.1.46 (KHTML, like Gecko) Version/9.0 Mobile/13D15 Safari/601.1

``` html
<paper-input type="date" label="no value (does not work)" />
<paper-input type="date" label="initial value (does work)" value="2016-03-03" />
```

https://plnkr.co/f5gzvmotEkEJ6zM5atsN

shawncplus:
Tested on iPhone 5S/iOS 11.0.2

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -237,6 +237,8 @@ This element is `display:block` by default, but you can set the `inline` attribu
         transition: transform 0.25s, width 0.25s;
         -webkit-transform-origin: left top;
         transform-origin: left top;
+        /* Fix for safari not focusing 0-height date/time inputs with -webkit-apperance: none; */
+        min-height: 1px;
 
         @apply --paper-font-common-nowrap;
         @apply --paper-font-subhead;


notwaldorf:
Focus/blur events are known to be red; can you provide a screenshot (or metrics or some on-the-record tesst) that this doesn't affect the size of the input/location of the underline etc

shawncplus:
This is about the only metrics/on-record-test I can think of https://imgur.com/a/V0zai

notwaldorf:
LGTM!

PR #587, opened 2017-11-07
352 fix applied to the right place this time

shawncplus:
Same same, but different See: https://github.com/PolymerElements/paper-input/issues/352#issuecomment-342626826


notwaldorf:
Do you have any screenshots/video to show that a) the sizing of this element isn't changing and b) it fixes the bug you want it to fix?

PR #594, opened 2017-11-21
fix label width in rtl; add :dir for polymer 2

### LINKED ISSUE #593: Paper input floating label reduces width in RTL
### Description
I want to use the paper-menu-dropdown with a floating label in a RTL document.

### Expected outcome
The floating label should span the whole width of the paper menu dropdown, both in LTR and RTL.

### Actual outcome
The floating label's width is reduced in RTL. 

### Live Demo
https://jsfiddle.net/5yw07fyt/

### Source
The culprit seems to be here?
https://github.com/PolymerElements/paper-input/blob/9a49d803c44bfec3e460495a22596ef2c233fd44/paper-input-container.html#L213
The width is set to 100% instead of 133%, scaling then effectively reduces the width to 75%.

### Browsers Affected
- [x] Chrome

notwaldorf:
- fixes #593: there was a TODO from like a year ago that made the effective width of the floated label in RTL smaller.
- Added `:dir` selectors so that this actually works in Polymer 2 with the polyfill (it doesn't on master)

There seems to be a separate bug where in Polymer 2 this still doesn't look perfectly right on Safari (the label is offset weirdly), but I am suspicious that's unrelated and a polyfill problem.

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -121,6 +121,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
       :host {
         display: block;
         padding: 8px 0;
+        overflow: hidden;
 
         --paper-input-container-shared-input-style: {
           position: relative; /* to make a stacking context */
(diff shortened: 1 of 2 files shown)


paper-input-container.html
```diff
@@ -259,12 +260,10 @@
         @apply --paper-input-container-label-floating;
       }
 
+      :host(:dir(rtl)) .input-content.label-is-floating ::slotted(label),
+      :host(:dir(rtl)) .input-content.label-is-floating ::slotted(.paper-input-label),
       :host-context([dir="rtl"]) .input-content.label-is-floating ::slotted(label),
       :host-context([dir="rtl"]) .input-content.label-is-floating ::slotted(.paper-input-label) {
-        /* TODO(noms): Figure out why leaving the width at 133% before the animation
```
e111077: 🍻

Issue #583, opened 2017-10-27
Styling for errorMessage and charCounter


stramel:
### Description

`errorMessage` and `charCounter` aren't accounted for in styling so they are often clipped due to other elements sitting underneath it.

### Expected outcome

When an `errorMessage` or `charCounter` are present they should be fully displayed when another element is directly underneath it.

### Actual outcome

When an `errorMessage` or `charCounter` are present they are clipped when another element is directly underneath it.

### Live Demo

![image](https://user-images.githubusercontent.com/855184/32121961-61208b20-bb24-11e7-84c8-7c6c3cc71b82.png)

![image](https://user-images.githubusercontent.com/855184/32121998-7d29e51e-bb24-11e7-92ba-1a33114aadc7.png)

### Browsers Affected

I believe all of these are affected but I have only tried Chrome.

- [x] Chrome
- [x] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

xelra:
I just ran into this problem myself. The funny thing is that it was working in my app and then suddenly I noticed the clipping just now.

Then I saw that even the demo is affected.
![image](https://user-images.githubusercontent.com/908568/33519752-b92cfb46-d7ad-11e7-95cf-98cb8f0386e0.png)

I can confirm that this also affects Firefox and Opera in addition to Chrome.

Does anyone have an intermediate fix? I've been playing with the CSS for some time now, but I just can't find the right values to change.

xelra:
I have identified the problem. `<paper-input-container>` is styled with `overflow: hidden`, which breaks the inheritance chain of visible overflow. Changing it to `overflow: inherit` or completely removing it from `:host {}` should fix the issue.

notwaldorf:
This bug was introduced in 2.0.4 (it was trying to fix something else,
sigh), and there’s a PR out to fix it! Will merge and do a release
tomorrow, hopeful. Sorry about that!

notwaldorf:
Closing as a dupe of #597, which has a better pinpoint to the actual error

PR #598, opened 2017-12-06
Fix overflow:hidden bug I introduced in the previous PR 

### LINKED ISSUE #593: Paper input floating label reduces width in RTL

### LINKED ISSUE #597: style issue in v2.0.4
### Description
The `paper-input` element display incorrently in inline-block mode!

### Expected outcome
screen shot 2017-12-05 at 12 30 56 pm

### Actual outcome
screen shot 2017-12-05 at 12 31 21 pm

### Live Demo
- [Demo of v2.0.3](https://jsbin.com/fonupur/3/edit?html,output)
- [Demo of v2.0.4](https://jsbin.com/fonupur/5/edit?html,output)

### Browsers Affected

- [x] Chrome

I don't check other browsers

notwaldorf:
Fixes #597, fixes #593 again.

Previous PR on this topic #594 did a bad (I did the bad). It added `overflow:hidden` which actually cut off the input and made it look like this:
screen shot 2017-12-05 at 4 54 38 pm

This PR correctly fixes the original RTL problem (which was that the label-come-down-from-floating-to-not-floating animation overflowed in the container) by correctly setting the overflow direction (which is backwards, thanks @bicknellr for spotting that)

Not floated (ltr and rtl):
screen shot 2017-12-05 at 5 35 37 pm

Floated:
screen shot 2017-12-05 at 5 35 43 pm

And as a bonus does the thing that @azakus said we should do which is just use `:dir` errwhere because it works in 1.0 as well.

It takes a village, fam.

### DIFF
--- a/paper-input-char-counter.html
+++ b/paper-input-char-counter.html
@@ -45,8 +45,7 @@ Custom property | Description | Default
         display: none !important;
       }
 
-      :host(:dir(rtl)),
-      :host-context([dir="rtl"]) {
+      :host(:dir(rtl)) {
         float: left;
       }
     </style>
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -259,9 +259,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
       }
 
       :host(:dir(rtl)) .input-content.label-is-floating ::slotted(label),
-      :host(:dir(rtl)) .input-content.label-is-floating ::slotted(.paper-input-label),
-      :host-context([dir="rtl"]) .input-content.label-is-floating ::slotted(label),
-      :host-context([dir="rtl"]) .input-content.label-is-floating ::slotted(.paper-input-label) {
+      :host(:dir(rtl)) .input-content.label-is-floating ::slotted(.paper-input-label) {
         right: 0;
         left: auto;
         -webkit-transform-origin: right top;
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -121,7 +121,6 @@ This element is `display:block` by default, but you can set the `inline` attribu
       :host {
         display: block;
         padding: 8px 0;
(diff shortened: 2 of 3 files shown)


xelra:
This also fixes #583. OMG! It's an issue-closing spree!

notwaldorf:
Yeah tests are the usual red but green locally; yay focus, yay blur.

AliMD:
@notwaldorf 
I teaching CSS more than 10 years but cant understand what relation between This PR file changed and #597 😕  😄 
Thank you, by the way, Could you please little explain what happened exactly ?! (if you like)

notwaldorf:
@AliMD In #594 I added an `overflow: hidden` to `paper-input-container` which clipped the content (like you saw, and worse, if you had an error message it would get clipped halfway). I originally did that because when in RTL, when the label needed to float back down (because you deleted all the content), it overflowed on the left hand side. That however, can be fixed by positioning the label correctly (the overflow direction gets flipped in RTL), so you don't need the `overflow:hidden`. Anyway, not a very scientific answer.

AliMD:
@notwaldorf Thank you but #597 not related to overflow, I change overflow in demo and problem not solved!

PR #602, opened 2017-12-19
[WIP] Unique IDs for input

### LINKED ISSUE #600: Error: non-unique id #input
### Description

`
`

It may be due to this?

### Steps to reproduce
2. Open the page in Chrome.

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

ipuiu:
Fixes #600 
Attempt to fix this by removing the id on the input.

Progress so far:
- [ ] Chrome: Some tests still fail (Only 4).
- [x] Firefox: All tests pass.

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -372,7 +372,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       _inputId: {
         type: String,
-        value: ''
+        value: function () {
+          return 'input_' + ( Math.random().toString(36).substring(2) );
+        }
       }
 
     },
@@ -393,7 +395,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * Returns a reference to the input element.
      */
     get inputElement() {
-      return this.querySelector('input');
+      return this.querySelector('.input-element');
     },
 
     /**
@@ -412,7 +414,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
     attached: function() {
       this._updateAriaLabelledBy();
-      this._updateInputId();
 
       if (this.inputElement &&
           this._typesThatHaveText.indexOf(this.inputElement.type) !== -1) {
@@ -523,10 +524,6 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
         label.id = labelledBy;
       }
       this._ariaLabelledBy = labelledBy;
-    },
-
-    _updateInputId: function () {
-      this._inputId = 'input_' + Polymer.PaperInputHelper.NextLabelID;
     },
 
     _onChange:function(event) {
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -19,6 +19,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
   Polymer.PaperInputHelper = {};
   Polymer.PaperInputHelper.NextLabelID = 1;
   Polymer.PaperInputHelper.NextAddonID = 1;
+  Polymer.PaperInputHelper.NextInputID = 1;
 
   /**
    * Use `Polymer.PaperInputBehavior` to implement inputs with `<paper-input-container>`. This
@@ -368,6 +369,11 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
       _ariaLabelledBy: {
         type: String,
         value: ''
+      },
+
+      _inputId: {
+        type: String,
+        value: ''
       }
 
     },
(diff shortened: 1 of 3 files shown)


ipuiu:
I signed it!

gitact:
Hello! Tried this solution too, the best one, but it seems they are only partial and limited solutions for input problems and derived polymers :-(

You know we surely cannot go to Production environment with our webapp logging twinklinkg Errors on Chrome 63 console, like lights on a Christmas tree :-P

_Any other brilliant (and final) ideas_ as Christmas gift from mythical Monica @notwaldorf and fascinating @robdodson (or vice versa) or other **Polymer Team** guys  ?
Thanks and greetings!

ipuiu:
Hi ! I'm aware of the tests failing for chrome 63 :( ... I won't give up though on finding a solution that will actually pass all the tests. I'll give it another try in the following days.

kornalius:
Any updates on this issue? The errors in Chrome 63 are causing problems for our QA department with their automated testing.

What is the best solution to fix the errors for input fields?

paper-input-behavior.html
```diff
@@ -388,7 +388,7 @@
      * Returns a reference to the input element.
      */
     get inputElement() {
-      return this.$.input;
+      return this.querySelector('.input-element');
```
notwaldorf: We should try to cache this -- the benefit of `this.$` is that it's cached, and you don't do this query all the time.

Something like this should work:
```
if (!this._inputElement) { 
  this._inputElement = this.querySelector('.input-element');
}
return this._inputElement;
```

paper-input.html
```diff
@@ -120,9 +120,10 @@
 
       <content select="[prefix]"></content>
 
-      <label hidden$="[[!label]]" aria-hidden="true" for="input">[[label]]</label>
+      <label hidden$="[[!label]]" aria-hidden="true">[[label]]</label>
```
notwaldorf: woof, this is bad. This break a11y. @robdodson, halp, do you have any suggestions?

paper-input-behavior.html
notwaldorf: (can we just do `this.querySelector('input');` and not add the extra useless class?>

notwaldorf: Actually: can we just assign to `this.$` rather than something else? So:

```
if (!this.$) { 
  this.$ = {}
}
if (!this.$.input) { 
  this.$.input = this.querySelector('input');
}
return this.$.input;
```

paper-textarea.html
```diff
@@ -131,15 +131,19 @@
...
-      this.$.input.textarea.setAttribute('aria-describedby', ariaDescribedBy);
+      this.textareaElement.setAttribute('aria-describedby', ariaDescribedBy);
     },
 
     get _focusableElement() {
-      return this.$.input.textarea;
+      return this.textareaElement;
     },
+
+    get textareaElement() {
```
notwaldorf: This shooooould be ok if we populate `this.$` correctly, right?

ipuiu:
@notwaldorf Many thanks for the feedback ! I'll gave those changes a try today !

paper-input.html
notwaldorf: We should probably just generate the ID using something like this: https://github.com/PolymerElements/paper-input/blob/master/paper-input-behavior.html#L21

notwaldorf: (and then manually add it to the label)

paper-textarea.html
ipuiu: Sadly this is actually the biggest issue I have so far. This is the main reason the tests fail ... the `paper-textarea`.

gitact:
It seems that despite the good will @notwaldorf and @ipuiu, we are still quite far from a solution of the Error logging :-(

paper-textarea.html
notwaldorf: Do you have any more details as to what fails? Are you setting the this.$ correctly?
I can check this out locally if you want

notwaldorf:
@ipuiu I figured out why your things were failing. I've pulled your commits into https://github.com/PolymerElements/paper-input/pull/608 (since I couldn't add a commit to this PR). Wanna take a look at that one instead? The tests seem to be green now :)

ipuiu:
@notwaldorf Many, many ... many thanks ! :D I didn't had all the time I wished to investigate properly the issue.

gitact:
@notwaldorf thank U too much. Always our _favourite developer's favourite developer_. Not only in our heart, but on the field! ;-P Thx

PR #607, opened 2018-01-17
Style mixin for input-wrapper

ankitsinghal0880:
I have a requirement to display paper-dropdown-menu as a standard dropdown and not in material design. I need to do it through stylesheet as few of our customer still want to have old look/feel and we do not want to have custom coding instead trying to change the look/feel using styles. I am able to display paper-input like legacy elements easily by using paper-input-container styles but no luck with dropdown.

Styling input wrapper will provide enterprise application to serve the customer needs without changing the underline code.

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -216,6 +216,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
       .input-wrapper {
         @apply --layout-horizontal;
         @apply --layout-center;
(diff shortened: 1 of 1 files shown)


ankitsinghal0880:
I signed it.

notwaldorf:
First, I can't tell anything from the diff since it looks like you've replaced the whole file? Secondly, do you have an example of exactly what you want to achieve, and why you can't use the existing mixins?

PR #608, opened 2018-01-18
1.x: Generate unique IDs for the input in the shadow root

### LINKED ISSUE #600: Error: non-unique id #input

notwaldorf:
The 2.x PR is here: https://github.com/PolymerElements/paper-input/pull/609

Fixes #600: Chrome has recently started throwing some auditing errors in the console if any `<input type="password">` has the same `id` as any other `input`. This means that in Shady DOM, if you have two `paper-inputs` on a page, you'd get this error. This fixes that by generating unique IDs for the inputs in the page.

(this is PR https://github.com/PolymerElements/paper-input/pull/602 with like one extra commit on it, thanks @ipuiu for doing all the work!)

### DIFF
--- a/paper-input-behavior.html
+++ b/paper-input-behavior.html
@@ -372,7 +372,9 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
 
       _inputId: {
         type: String,
-        value: ''
+        value: function () {
+          return 'input_' + ( Math.random().toString(36).substring(2) );
+        }
       }
 
     },
@@ -393,7 +395,7 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
      * Returns a reference to the input element.
      */
     get inputElement() {
-      return this.querySelector('input');
+      return this.querySelector('.input-element');
     },
 
     /**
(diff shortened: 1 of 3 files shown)


notwaldorf:
(Working on the 2.x PR for this same problem; it's a bit more complicated because of timing)

paper-input-behavior.html
```diff
@@ -532,8 +532,10 @@
       this._ariaLabelledBy = labelledBy;
     },
 
-    _updateInputId: function() {
-      this._inputId =  'input-' + Polymer.PaperInputHelper.NextInputID++;
+    _generateInputId: function() {
+      if (!this._inputId || this._inputId === '') {
```
ipuiu: @notwaldorf Maybe I'm missing the big picture ... ` if (!this._inputId)` wouldn't be enough ?

notwaldorf: i was being paranoid :D

PR #609, opened 2018-01-18
2.x: Generate unique IDs for the input in the shadow root

### LINKED ISSUE #600: Error: non-unique id #input

notwaldorf:
The 1.x PR is here: https://github.com/PolymerElements/paper-input/pull/608

Fixes #600: Chrome has recently started throwing some auditing errors in the console if any `<input type="password">` has the same `id` as any other `input`. This means that in Shady DOM, if you have two `paper-inputs` on a page, you'd get this error. This fixes that by generating unique IDs for the inputs in the page.

### DIFF
--- a/paper-input.html
+++ b/paper-input.html
@@ -230,7 +230,7 @@ Custom property | Description | Default
         allowed-pattern="[[allowedPattern]]"
         invalid="{{invalid}}"
         validator="[[validator]]">
-      <input id="nativeInput"
+      <input
         aria-labelledby$="[[_ariaLabelledBy]]"
         aria-describedby$="[[_ariaDescribedBy]]"
         disabled$="[[disabled]]"
@@ -307,7 +307,7 @@ Custom property | Description | Default
 
     _onIronInputReady: function() {
       if (this.inputElement &&
-          this._typesThatHaveText.indexOf(this.$.nativeInput.type) !== -1) {
+          this._typesThatHaveText.indexOf(this.$$('input').type) !== -1) {
         this.alwaysFloatLabel = true;
       }
 
--- a/paper-input.html
+++ b/paper-input.html
@@ -306,8 +306,13 @@ Custom property | Description | Default
     },
 
     _onIronInputReady: function() {
+      // Even though this is only used in the next line, save this for
+      // backwards compatibility, since the native input had this ID until 2.0.5.
+      if (!this.$.nativeInput) {
+        this.$.nativeInput = this.$$('input');
+      }
       if (this.inputElement &&
-          this._typesThatHaveText.indexOf(this.$$('input').type) !== -1) {
+          this._typesThatHaveText.indexOf(this.$.nativeInput.type) !== -1) {
         this.alwaysFloatLabel = true;
       }
 
--- a/paper-input.html
+++ b/paper-input.html
@@ -161,7 +161,7 @@ Custom property | Description | Default
 
       <slot name="prefix" slot="prefix"></slot>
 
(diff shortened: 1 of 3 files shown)


paper-input.html
```diff
@@ -221,7 +223,9 @@
 
   <template id="v1">
     <!-- Need to bind maxlength so that the paper-input-char-counter works correctly -->
-    <iron-input bind-value="{{value}}" id="input" slot="input"
+    <iron-input bind-value="{{value}}" slot="input"
+        class="input-element"
+        id$="[[_inputId]]"
```
valdrinkoshi: should this be done in the native input below?

notwaldorf: no, i fyou look, the `id` was in the `iron-input` previously

valdrinkoshi: Chrome complains about native inputs having the same id, e.g. this is the simplified DOM when we're in shadyDOM mode, and Chrome complains about double `#nativeInput`:
```html
  <form>

    <label for="p0">password p0</label>
    <iron-input id="p0">
      <input type="password" id="nativeInput">  
    </iron-input>

    <label for="p1">password p1</label>
    <iron-input id="p1">
      <input type="password" id="nativeInput">  
    </iron-input>

  </form>
```
http://jsbin.com/gelevij/1/edit?html,console,output

Here the repro with paper-input having the same issue:
```html
  <form>

    <paper-input type="password" label="password p0"></paper-input>

    <paper-input type="password" label="password p1"></paper-input>
    
  </form>
```
http://jsbin.com/puyiwiw/1/edit?html
http://output.jsbin.com/puyiwiw/?wc-ce=true&wc-shadydom=true

notwaldorf: oh i see (yes: the error in Chrome is what we're trying to fix btw). I guess `nativeInput` also needs a separate ID (paper-input doesn't know if you're stamping a 1.x or 2.x template, so that's why things have to be named similarly)

valdrinkoshi: I think we only need the fix for `nativeInput`, not `<iron-input>`

notwaldorf: Sure, but `inputElement` the getter has to point to the right thing (either the native input in 1.x or the `iron-input` in 2.x, and since it's the same code, it also has to have a correctly generated name (so that the getter works).

notwaldorf: I can just remove the `nativeInput` id here, tho, since it's only used in one place

notwaldorf:
@valdrinkoshi comment addressed, PTAL?

paper-input.html
```diff
@@ -303,7 +307,7 @@
 
     _onIronInputReady: function() {
       if (this.inputElement &&
-          this._typesThatHaveText.indexOf(this.$.nativeInput.type) !== -1) {
+          this._typesThatHaveText.indexOf(this.$$('input').type) !== -1) {
```
valdrinkoshi: instead of doing this, WDYT of setting `this.$.nativeInput = this.$$('input')` on `ready`? I'm surely over-thinking this, but there might be cases where people rely on these internal ids...

paper-input.html
valdrinkoshi: Nice call! 👍

notwaldorf:
🍏 📗 💚 🎉

Issue #605, opened 2018-01-08
paper-input title stuck after reset


punkratz312:
The paper-input does not rerender itself after resetting its value to empty. The label gets cleared as expected but the title label does not move back ontop the input.

![image](https://user-images.githubusercontent.com/8830888/34677182-9219776a-f48f-11e7-9d17-7c0f824da05b.png)

punkratz312:
@notwaldorf  plz reply

notwaldorf:
I can't repro anymore (I think this was fixed in the recent release): http://jsbin.com/botetem/edit?html,output

punkratz312:
The bug is still active!
In your example you reset the value by explicit hand of course this will work.
In my case and thats more like a real world scenario i have a binding to the elements value. The binding works, the input is empty, but still does not rerender it selfe.

notwaldorf:
Then do you have some repro steps? You didn't provide any, so I can't really debug anything from just a screenshot.

punkratz312:
I´ve tried to define a custom element in the js bin but didnt made it work, sorry.

Just define an working custom element  that has one property, bind the input´s value to the prop and clear the property after a timeout. The input will be cleared but not rerendered like in the screenshot.

punkratz312:
@notwaldorf plz try to reproduce the bug, I have no idea how to define a custom polymer element in the js bin. Binding the value to a property and resetting it will lead to this very obviouse bug.

notwaldorf:
Here's a repro step: http://jsbin.com/lirifuy/edit?html,output

The problem seems to be when you set the value to `undefined`. It seems fine with `null`. As a workaround while looking for a fix, can you make sure you're setting the value to `null`, maybe?

punkratz312:
awesome, thx.

punkratz312:
@notwaldorf nice thanks for the fix. When are you going to release it?

PR #615, opened 2018-02-01
Fix label not resetting on undefined

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/605


notwaldorf:
hold on, reopening as a new PR because something's gone wonky

PR #616, opened 2018-02-01
Fix label not resetting on undefined

### LINKED ISSUE #605: paper-input title stuck after reset

notwaldorf:
Fixes #605

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -563,13 +563,11 @@ This element is `display:block` by default, but you can set the `inline` attribu
       // the rest of the time (specifically: it does not validate it at startup, but
       // it does after that. If this is in fact the bootup case, ignore validation,
       // just this once.
-      if (this.__isFirstValueUpdate && input.value === undefined) {
-        this.__isFirstValueUpdate = false;
-        return;
-      }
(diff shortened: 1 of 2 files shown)


paper-input-container.html
```diff
@@ -553,17 +559,18 @@
...
-      // case  we care about manifests itself when the value is undefined.
-      // If this causes future problems, we need to do something like track whether
-      // the iron-input-ready event has fired, and this handler came before that.
-
-      if (input.value === undefined) {
+      // Paper-input treats a value of undefined differently at startup than
+      // the rest of the time (specifically: it does not validate it at startup, but
+      // it does after that. If this is in fact the bootup case, ignore validation,
+      // just this once.
+      if (this.__isFirstValueUpdate && input.value === undefined) {
```
valdrinkoshi: nit: WDYT of wrapping the previous if condition?
```js
if (this.__isFirstValueUpdate) {
  this.__isFirstValueUpdate = false;
  if (input.value === undefined) {
    return;
  }
}

notwaldorf: Very pro this change, thanks

PR #618, opened 2018-02-12
Added missed slot attributes

stramel:
- Added missed slot attributes


notwaldorf:
The non-travis changes LGTM, I have no idea about the travis config so I defer to @azakus

stramel:
@notwaldorf NVM, something was being wonky... Haven't tested those changes with headless browsers in containerized. Will put out separately.

notwaldorf:
in that case :shipit:

PR #623, opened 2018-02-26
Added mixin for calendar picker indicator for date type input

ConorPender:
This change has already been merged into master, but it is also needed on 1.x.


notwaldorf:
Thanks!

Issue #621, opened 2018-02-23
Cannot style paper-input-container's input correctly


kevinpschaaf:
### Description
The docs for `paper-input-container` indicate:
> You can style the nested `<input>` however you want; if you want it to look like a Material Design input, you can style it with the --paper-input-container-shared-input-style mixin.

However, due to limitations in the @apply shim, the `--paper-input-container-shared-input-style` is not available to be applied from outside the element; as such, styling the input as described in the docs does not work:

```html
<custom-style>
  <style>
    input {
      @apply --paper-input-container-shared-input-style;
    }
  </style>
</custom-style>

<paper-input-container>
  <iron-input slot="input">
    <input>
  </iron-input>
</paper-input-container>
```

### Expected outcome
The input looks like a `paper-input`

### Actual outcome
The input is unstyled

### Live Demo
http://jsbin.com/loketi/edit?html,output

### Steps to reproduce
1. Open jsBin
2. Note input is ugly
3. Uncomment custom style with outer definition of `--paper-input-container-shared-input-style` and note input is nice

### Browsers Affected
All

PR #625, opened 2018-02-28
fix the shared-input-style which was in the wrong spot

notwaldorf:
Fixes https://github.com/PolymerElements/paper-input/issues/621

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -116,30 +116,35 @@ This element is `display:block` by default, but you can set the `inline` attribu
 `display:inline-block`.
 -->
 
+<custom-style>
+  <style is="custom-style">
+    html {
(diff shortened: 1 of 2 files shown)


paper-input-container.html
```diff
@@ -116,30 +116,35 @@
 `display:inline-block`.
 -->
 
+<custom-style>
+  <style is="custom-style">
+    html {
```
valdrinkoshi: In case we want to backport it to 1.x, we might need to add `:root` here

notwaldorf: i'll add it just in case

notwaldorf: oh tldr not needed (this is a 1.x PR, let's cross that bridge when we get there)

PR #628, opened 2018-03-07
Expose mixins to configure .input-content

### LINKED ISSUE #535: allow flex-wrap styling of input-content div
### Description
We have a case where we have several longer prefix elements, unfortunately we can't find a way to style the input-content div so that it wraps the prefix elements. Would it be possible to add a mixin to that div so that it can be styled? Other Idea?

victorbalan:
Fixes #535 
Add `--paper-input-container-input-content` and `--paper-input-container-input-content-focus` mixins.


victorbalan:
@notwaldorf Seems like you have the most commits. Would you please review and maybe push it forward towards the 2.x branch if it respects the guidelines?

victorbalan:
It not only fixes the mentioned issue but also blocks the upgrade to polymer 2.x in my case because I used the `::shadow` selector to apply custom styling on the `.input-content` div.

Issue #636, opened 2018-03-23
paper-input-container-input font-size: inherit no longer works.


freshp86:
When using the latest `iron-input` (which wraps a native `<input>`), font-size inherit no longer works inside the `paper-input-container-input` mixin, because it instead inherits from the parent `iron-input`. Minimal example below:

```html
<!DOCTYPE html>                                                                                                                                                                                                     
<html>
  <head>
    <script>
       Polymer = {
         dom: 'shadow',
         lazyRegister: true,
         suppressBindingNotifications: true,
         suppressTemplateNotifications: true,
         useNativeCSSProperties: true,
       };
    </script>
    <link href="polymer/polymer.html" rel="import">
    <link href="paper-input/paper-input.html" rel="import">
  </head>
  <body>
    <style>
      body {
        font-size: 10px;
      }   
    </style>
    <dom-module id="x-example">
      <template>
       <style>
         paper-input {
           --paper-input-container-input: {
             font-size: inherit;
           };
         }
       </style>
       <paper-input value="foo"></paper-input>
      </template>
    </dom-module>

    <div>I am 10px</div>
    <x-example></x-example>
    <script>
      Polymer({is: 'x-example'});
    </script>
  </body>
</html>
```

Before versions:
paper-input: 2.1.0
iron-input: 1.0.10
Observing <input is="iron-input"> inheriting 10px font from body.

After versions:
paper-input: 2.2.0
iron-input: 2.1.2
Observing inner <input> inheriting 16px from it's parent <iron-input>.

### Proposed fix?
See line https://github.com/PolymerElements/paper-input/blob/master/paper-input-container.html#L293 and compare with lines 300-301. Shouldn't both mixins also be applied on the iron-input, such that iron-input does not end up with a 16px font size?

cc @notwaldorf

feuerste:
I agree, the `--paper-input-container-input` mixin should be applied to `.input-content ::slotted(iron-input)` as well.

notwaldorf:
Yeah, I think this is also needed for other mixins, like https://github.com/PolymerElements/paper-input/blob/master/paper-input-container.html#L309

Issue #613, opened 2018-01-29
Paper input container mixin not applying


watsab:
Hi!
I've got an issue with a mixin that is not applied and I guess it's linked to a css rule that may need some change.

Here's the style:
```
:host {
  --input-border: {
    -webkit-border-radius: 1px;
    -moz-border-radius: 1px;
    border-radius: 1px;
    border: 1px solid var(--input-border-color);
    box-sizing: border-box;
  };

  --error-color: #ec2c45;
  --focus-color: #4296b4;
}

paper-input  {
  --paper-input-container-input: {
      font-size: var(--input-font-size);
      display: block;
      background-color: #ffffff;
      color: var(--primary-text-color);
      height: var(--input-height);
      padding-left: 2rem;
      @apply --input-border;
  };
  --paper-input-container-input-focus: {
    border-color: var(--focus-color);
  };
  --paper-input-container-input-invalid: {
    border-color: var(--error-color);
  };
}
```

The `--paper-input-container-input` is actually applied but when the input is invalid or focused, the border color is not applied as I would expect.
My guess is because of the following css rule:
```
.input-content.focused ::slotted(input),
.input-content.focused ::slotted(textarea),
.input-content.focused ::slotted(iron-autogrow-textarea),
.input-content.focused ::slotted(.paper-input-input) {
  @apply --paper-input-container-input-focus;
}
```

In v2 the input is actually surrounded by an iron-input, and that would be why the mixin is not applied although I don't understand why the mixin `--paper-input-container-input` is applied in the first place and not `--paper-input-container-input-focus` or `--paper-input-container-input-invalid`.

Thanks!

notwaldorf:
Fixed in https://github.com/PolymerElements/paper-input/pull/639

PR #639, opened 2018-03-28
styles not applied to slotted input in 2.x

### LINKED ISSUE #638: Unable to style paper input focus
### Description

I want to style paper input by changing the focus underline to a full box border with a dropshadow. I can't seem to find out which mixin is supposed to change the underline to a full box border.  Tried all mixins from paper-input-container with no succes.

### Expected outcome

On focus give the paper input a full border. (see screenshot)

screen shot 2018-03-26 at 14 09 12

### Actual outcome

Only underline is visible, able to change its colour and state (solid, etc)

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari 9
- [x] Safari 8
- [x] Safari 7
- [x] Edge
- [x] IE 11
- [x] IE 10

notwaldorf:
Fixes #638, #636, #613.

It's hard to add a test for this, so I've added a super complex `demo-snippet`:
screen shot 2018-03-28 at 12 32 20 pm

### DIFF
--- a/paper-input.html
+++ b/paper-input.html
@@ -91,13 +91,21 @@ Custom property | Description | Default
       input {
         /* Firefox sets a min-width on the input, which can cause layout issues */
         min-width: 0;
-        @apply --paper-input-container-shared-input-style;
-        @apply --paper-input-container-input;
       }
 
-      input::-webkit-outer-spin-button,
-      input::-webkit-inner-spin-button {
-        @apply --paper-input-container-input-webkit-spinner;
+      /* In 1.x, the <input> is distributed to paper-input-container, which styles it.
+      In 2.x the <iron-input> is distributed to paper-input-container, which styles
+      it, but in order for this to work correctly, we need to reset some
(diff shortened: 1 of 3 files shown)


paper-input.html
```diff
@@ -89,27 +89,17 @@
...
-
         /* Firefox sets a min-width on the input, which can cause layout issues */
         min-width: 0;
-
-        @apply --paper-font-subhead;
+        @apply --paper-input-container-shared-input-style;
         @apply --paper-input-container-input;
       }
 
+      input::-webkit-outer-spin-button,
```
valdrinkoshi: looks like a duplicate, same rules applied few lines below

paper-input.html
```diff
@@ -89,27 +89,17 @@
       }
 
       input {
-        position: relative; /* to make a stacking context */
```
valdrinkoshi: 💯 for removing this block & reusing the mixin!

Issue #606, opened 2018-01-16
Styling paper-dropdown-menu


ankitsinghal0880:
I have a requirement to display paper-dropdown-menu as a standard dropdown and not in material design. I need to do it through stylesheet as few of our customer still want to have old look/feel and we do not want to have custom coding instead trying to change the look/feel using styles. I am able to display paper-input like legacy elements easily by using paper-input-container styles but no luck with dropdown.

I need input wrapper to have a border. Currently this isn't possible.

Proposal for a new mixin- 

.input-wrapper {
@apply --layout-horizontal;
@apply --layout-center;
**@apply --paper-input-container-input-wrapper This new mixin is required.**
position: relative;
}

### Browsers Affected

- [ ] Chrome
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

AlbertoFdzM:
I have this problem too.

Although I think it will be better to place the mixin at the end of the CSS block to be able to override all previous rules.

PR #657, opened 2018-06-04
Add mixin on input-wrapper

RoXuS:
Add a mixin on input-wrapper


AlbertoFdzM:
This PR should fix #606

PR #658, opened 2018-06-06
Fixes Auto-Validation fires directly for required without any user interaction

aarondrabeck:
This resolves issues [134](https://github.com/PolymerElements/iron-input/issues/134) and [655](https://github.com/PolymerElements/paper-input/issues/655) introduced from [PR 130](https://github.com/PolymerElements/iron-input/pull/130).

Allowing '' here (along with undefined) should not create any issues because it is only checking once on startup.  The alternative is to revert PR 130 but I am not sure why setting a default was needed.

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -556,7 +556,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
       // validation, just this once.
       if (this.__isFirstValueUpdate) {
         this.__isFirstValueUpdate = false;
-        if (input.value === undefined) {
+        if (input.value === undefined || input.value === '') {
           return;
         }
       }


e111077:
Come think of it, why would the default value ever be `undefined` with the change in PolymerElements/iron-input#130? In polymer 1 it is a native `input` that is extended so it would be `''` and polymer 2 we set it to `''` in `iron-input`

e111077:
A test would also be helpful to make sure this regression doesn't happen again

aarondrabeck:
@notwaldorf  Thanks for the merge and test!  Is there any way we could get a release with this in it?

e111077:
We are currently reviewing our internal test suite results. You should expect a release later today or a rollback depending on the result

aarondrabeck:
Thanks!

PR #660, opened 2018-06-19
Change ts type to any

e111077:
related PolymerElements/iron-form-element-behavior#47

This is because one often extends both this behavior and iron form element behavior at the same time. This causes TS issues when their types diverge


e111077:
do not merge, pending internal tests

e111077:
tests are 👍

PR #682, opened 2018-11-13
Skip a 'blur' event test if the browser has a bug that would prevent it from working.

bicknellr:
Firefox seems to have a bug where calling `blur` on an element with a ShadowRoot containing a focused element doesn't do anything. This checks for the bug and skips the test if present.


test/paper-input.html
```diff
@@ -404,6 +404,27 @@
           });
 
       test('blur events fired on host element', function(done) {
+        // Firefox's Shadow DOM implementation seems to have a bug where
+        // calling `blur` on an element with a shadow root containing the true
+        // focused node doesn't blur the node.
+        const shouldSkip = (() => {
```
e111077: Why use an iife? Please remove if exraneous.

Also please clean up DOM after check.

PR #691, opened 2019-03-04
make sure event.target is the real addons target

christophe-g:
We want the real addons target to be added as addons. 

Due to shadowDom event retargeting, `event.target` is is not the addon anymore in some advanced templating scenario. As a result, `updateAddons` call a not-existing `update` method, creating an error. 

This PR makes sure we add the initial event target.

This PR is against the 2.x branch (which I use), but the observed error is expected to happen under 3.x as well.

### DIFF
--- a/paper-input-container.html
+++ b/paper-input-container.html
@@ -521,7 +521,7 @@ This element is `display:block` by default, but you can set the `inline` attribu
       if (!this._addons) {
         this._addons = [];
       }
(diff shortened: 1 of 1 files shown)


e111077:
Unfortunately we are still doing a large internal migration that may result in an external major release of paper-input. We are in a code freeze for the element and iron-input in the meanwhile.

PR #692, opened 2019-03-13
Prevent error messages from being read by screen readers when the input is not in an error state.

bicknellr:
Hide the content of paper-input-error elements with `invalid === false` using `visibility: hidden;` on an internal wrapper element so that it isn't read by screen readers even if the paper-input-error is referenced by an `aria-describedby` attribute.

### DIFF
--- a/paper-input-error.js
+++ b/paper-input-error.js
@@ -55,10 +55,29 @@ Polymer({
 
       :host([invalid]) {
         visibility: visible;
-      };
+      }
+
+      #a11yWrapper {
+        visibility: hidden;
+      }
+
+      :host([invalid]) #a11yWrapper {
+        visibility: visible;
+      }
     </style>
 
-    <slot></slot>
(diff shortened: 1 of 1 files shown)


e111077:
I just wonder how this will work when people use paper-input-error on their own and styling it (e.g. will we have to inherit certain styles?). I don't think there will be any issues, but we'll see from internal testing.

bicknellr:
Internal testing ended up clean after a deflake. (Links to the tests are in the CL you've already seen: http://go/polymerelements_paper-input_692_cl)

bicknellr:
@e111077 btw that link is still available in the history of the comment. I think if you want to remove it entirely you'll have to delete the comment itself. FWIW, I don't think it's a problem to post go/ links: https://chromium.googlesource.com/chromium/src/+/master/docs/mac_build_instructions.md

PR #701, opened 2019-12-06
Add properties to allow setting the internal input's ARIA role and aria-haspopup attribute.

bicknellr:
This is part of PolymerElements/paper-dropdown-menu#317, which needs to set these to be able to continue using paper-input for the sake of styling but also allow paper-input's wrapped input to act as the focusable element while being read by screen readers as a button rather than a text input.


bicknellr:
Here be dragons. I'll try a test import before merging just to see if anything explodes.

